Issue #40, opened 2017-08-28
Tokens File


jamesmelzer:
### Deliver

* Place YAML tokens file in site hierarchy
* Setup build process to convert YAML to JSON to SASS
* Create YAML hierarchy

see Design #39

PR #149, opened 2018-06-15
#40 tokens file

jamesmelzer:
Dropping this into PR early to get feedback and comments. I expect to iterate on this before merging.

### DIFF
--- a/library/tokens/tokens.yaml
+++ b/library/tokens/tokens.yaml
@@ -31,10 +31,6 @@ text-color :
   secondary : *color-neutral-35
   link :      *color-blue-default
   disabled :  *color-neutral-70
-background-color :
-  shell :     *color-white
-  card :      *color-neutral-95
-  opening :   *color-neutral-90
 icon-color :  *color-blue-default
 interactive-color :
   focus:      *color-blue-30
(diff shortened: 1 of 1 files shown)


library/tokens/tokens.yaml
```diff
@@ -0,0 +1,183 @@
+## spirit-design-system v0.0.1
+##--------------------------------------------------
+# ====== palette ======= #
+color :
+  black : &color-black      "#00081A"
+  neutral :
+    35  : &color-neutral-35 "#575859"
+    70  : &color-neutral-70 "#AFB0B3"
+    95  : &color-neutral-95 "#536065"
```
jamesmelzer: typo: should be #EDEFF2

joesnellpdx:
@jamesmelzer I LOVE the use of REM vs PX, but we will need to have documentation on setting the root default text size on with the consuming website / application - or the styles will be altered (i.e. if root is 14px instead of 16px - styles will be smaller ).

If we are setting that recommendation, one idea is to set the root to 10px so 1rem = 10px for simple math + variable usage. This may be too much to ask for consuming applications / sites, but worth pointing out at this time. I'm not suggesting a change - more of an FYI.

CC: @kevinmpowell

PR #150, opened 2018-06-19
update typography and add breakpoints

jamesmelzer:
- added breakpoint values
- updated type line-heights
- corrected formatting in type section

**Review question: Any concerns with using pixels for line-heights?** I expressed heading line-heights in pixels, since they're all unique and it required many decimal places or rounding to express them in proportions.

### DIFF
--- a/library/tokens/tokens.yaml
+++ b/library/tokens/tokens.yaml
@@ -164,40 +164,44 @@ font :
     regular : 400
     bold : 600
   size :
-    base: &font-size-base 16px
+    base : &font-size-base 16px
     xs : 13px
     s :  14px
     m :  16px
     l :  18px
-    subhead:                      16px
-    subhead-largescreen:          24px
-    display-level-1:              56px
-    display-level-1-largescreen:  96px
-    display-level-2:              42px
-    display-level-2-largescreen:  72px
-    heading-level-1:              32px
-    heading-level-1-largescreen:  56px
-    heading-level-2:              24px
-    heading-level-2-largescreen:  44px
-    heading-level-3:              18px
-    heading-level-3-largescreen:  32px
-    heading-level-4:              16px
-    heading-level-4-largescreen:  24px
-    heading-level-5:              14px
-    heading-level-5-largescreen:  18px
-    heading-level-6:              13px
-    heading-level-6-largescreen:  16px
-    code: 13px
-  line-height:
-    body: 1.5          # check this in Sketch
-    prose: 1.5         # check this in Sketch
-    controls: 1
-    display-level-1: 1.25 # reload the branch before doing these, check mobile sizes too
-    display-level-2: 1.25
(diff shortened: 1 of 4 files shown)


library/tokens/tokens.yaml
```diff
@@ -164,40 +164,44 @@ font :
     regular : 400
     bold : 600
   size :
-    base: &font-size-base 16px
+    base : &font-size-base 16px
```
kevinmpowell: If we're not using `rems` drop this. It's not needed and will get confused with `font-size-m`

library/tokens/tokens.yaml
```diff
@@ -164,40 +164,44 @@ font :
...
-    heading-level-2: 1.25
-    heading-level-3: 1.25
-    heading-level-4: 1.25
-    heading-level-5: 1.25
-    heading-level-6: 1.25
-
+    subhead :                      16px
+    subhead-largescreen :          24px
+    display-level-1 :              56px
+    display-level-1-largescreen :  96px
```
kevinmpowell: I'd prefer mapping `largescreen` to a specific breakpoint if we know when it's going to switch. `display-level-1-breakpoint-l` will make it clearer how to assemble this functionality in the tokens.

library/tokens/tokens.yaml
```diff
@@ -164,40 +164,44 @@ font :
...
+    heading-level-4-largescreen :  24px
+    heading-level-5 :              14px
+    heading-level-5-largescreen :  18px
+    heading-level-6 :              13px
+    heading-level-6-largescreen :  16px
+    code : 13px
+  line-height :
+    body : 1.5    # use for prose and anything with paragraphs
+    controls : 1  # use inside components for labels and microcopy
+    display-level-1 : 108px
```
kevinmpowell: We need to convert these to unitless values for more consistent rendering. `px` based line heights can be problematic.

library/tokens/tokens.yaml
```diff
@@ -164,40 +164,44 @@ font :
...
+    display-level-2 : 76px
+    heading-level-1 : 72px
+    heading-level-2 : 56px
+    heading-level-3 : 40px
+    heading-level-4 : 32px
+    heading-level-5 : 24px
+    heading-level-6 : 20px
+breakpoint : 
+  xs : 0      #6 col auto
+  s  : 375px  #6 col 343px
```
kevinmpowell: I don't understand the comments. What's the difference between the token value and the value in comments?

joesnellpdx:
@jamesmelzer 

Default button on sketch:
screen shot 2018-06-20 at 11 36 24 am

jamesmelzer:
@joesnellpdx for now, include it as a variable in button. If Patrick or Ismael indicate we're reusing it elsewhere, we can promote it into tokens.

Issue #59, opened 2017-08-28
Button


jamesmelzer:
### Deliver

* Kitchen sink file showing all variations in all states, including too much and too little content
* Macro to create all variations
* Update tokens file with necessary style decisions

joesnellpdx:
@jamesmelzer @kevinmpowell 

How do I make changes / updates to tokens file?

I see "// DO NOT EDIT: This file is automatically generated by a build task", but adding to the tokens.yaml and tokens.json, but not working as expected.

kevinmpowell:
@joesnellpdx the `/tokens/tokens.yaml` file is where you should make your changes. As long as the main `gulp` task is running it should automatically update `tokens.scss` and `tokens.json`

joesnellpdx:
@jamesmelzer @kevinmpowell 

About ready to move back into peer review, but have some questions:

1. Is there a way to prefix svg icons? I added icons, but prefixing with 'ie'. I'm guessing there is a simple update to a gulp task.
2. How do I attach a commit to an issue ( like Kevin's above), or is this not a big deal?
3. I added a simple doc-site template, but not sure if that is part of this task. Should I continue with that?
4. I added fonts, but not able to accurately test text-crop mixin so line-height still at `1`
5. Border for focus is using inset box shadow per our discussion last week - going to come up again with borders on form inputs
6. I did not add any font loading enhancements - just `@font-face` for now ( see: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization )
7. Updated / added some tokens - please review to make sure my assumptions are correct
8. Nujucks macro - new to this - please review implementation - got complex 
9. Where to engineering / sketch workflow docks end up? Goal to put into doc-site or are they internal?
10. To clarify, all high level conversation on PRs, tickets, issues will be in the associated issue ( here )?

Sorry for the laundry list - LMK if you would rather I ask in a different forum / place.

Branch:
https://github.com/JDRF/spirit/pull/152/files

kevinmpowell:
@joesnellpdx this is the right place for these questions. I'll have time to respond this afternoon after our sprint planning.

kevinmpowell:
@joesnellpdx I'm going to move my responses over to the PR #152 so we consolidate the conversation

joesnellpdx:
@kevinmpowell

Thanks for answering questions! I believe I have addressed all your issues by my latest commits or by comment response.

[...]

PR #152, opened 2018-06-20
Feature/59 component button

joesnellpdx:
Added library component button

- Not able to add to doc-site locally
- Custom variables in buttons.scss: need discussion on tokens
- box sizing and border: not able to alter box sizing to have button ignore border width

For box sizing, options:

- Kevin passes knowledge to accomplish with box sizing
- Use pseudo element
- Use box shadow inset
- Use custom padding ( remove border width )

Testing process: 
LMK where / how workflow should be altered + any other feedback 
Also, not all buttons complete at this time, but want review on current state of branch.
I assume PRs and Issues will not be directly connected besides branch titles


joesnellpdx:
Doc site enabled. 
Continuing to add missing buttons.

library/components/button/button.njk
```diff
@@ -1,7 +1,15 @@
...
-          text="Button Text") %}
-  <button class="spirit-button{{ ' ' + class if class }}">
-    {{ text }}
-  </button>
+          disabled=false,
+          text="Button Text",
+		  tag="button",
+		  type=false,
+		  href=false) %}
+	{% if 'input' === tag  %}
```
kevinmpowell: Not sure if this is just a "work-in-progress (WIP)" artifact, but I'm not sure what value this `if` statement is adding. The only difference between the branches seems to be the ability to add a `disabled` attribute.

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
...
 .spirit-button {
-  background: blue;
-  border-radius: 5px;
-  border: 0;
-  color: white;
+  background: $spirit-interactive-color-primary-fill;
+  border: $spirit-border-button solid $spirit-interactive-color-primary-fill;
+  border-radius: $spirit-border-radius-m;
+  box-shadow: $spirit-elevation-1 $spirit-interactive-color-primary-glow;
+  box-sizing: border-box; // does not inset the border ignoring the padding ( can use inset shadow, pseudo element, or padding modifications )
```
kevinmpowell: This is what I expected. Let's revisit with @jdrf-digitaldesigner in a crit to see if this needs to be changed.

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
+// Letter Spacing
+$spirit-button-font-letter-spacing-15: .15em;
+
+// Transitions
+$spirit-button-transition-duration: .3s;
+$spirit-button-transition-timing-function: ease-out;
+
+// tokens.yaml not tranlating correct colors for blues to tokens.scss
+$spirit-interactive-color-focus: $spirit-color-blue-30;
```
kevinmpowell: Yes, this should definitely be a token.

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
+// Letter Spacing
+$spirit-button-font-letter-spacing-15: .15em;
```
kevinmpowell: If this is button specific then leave as-is. @jamesmelzer do you have a sense on other places a custom letter spacing is used?

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
...
+
+// Transitions
```
kevinmpowell: These basic hover transitions will likely become tokens. I can see these being applied to links, tabs, etc. @jamesmelzer thoughts?

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
...
-  font-size: 16px;
-  padding: 10px 20px;
+  display: inline-block;
+  font-family: $spirit-font-family-sans-serif;
+  font-size: $spirit-font-size-m;
+  font-style: normal;
+  font-weight: $spirit-font-weight-bold;
+  letter-spacing: $spirit-button-font-letter-spacing-15;
+  line-height: $spirit-font-line-height-controls;
+  margin: 0;
```
kevinmpowell: Delete line 28 and change line 27 to `margin: $spirit-space-stack-2-x;`

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
...
+  }
+
+  &:focus,
+  &:focus-within {
+    border: $spirit-border-button solid $spirit-interactive-color-focus;
+  }
+}
+
+// for testing - to be replaced with section component - need direction here
+.button-wrap-dark-test {
```
kevinmpowell: I'd go with something like `.spirit-container--dark` such that your markup is:
```
<div class="spirit-container--dark">
  <button class="spirit-button">
</div>
```

And the SCSS becomes:
```
.spirit-button {
  // styles here
  .spirit-container--dark & {
    //.spirit-button--on-dark styles
  }
}
```

This removes the need to add a modifier class to each individual component and allows a parent context to trigger the switch. This would replace the `.spirit-button--on-dark` modifier class.

library/components/button/button.njk
```diff
@@ -1,7 +1,15 @@
...
           class=false,
-          text="Button Text") %}
-  <button class="spirit-button{{ ' ' + class if class }}">
-    {{ text }}
-  </button>
+          disabled=false,
+          text="Button Text",
+		  tag="button",
+		  type=false,
+		  href=false) %}
```
kevinmpowell: Allow `href` to automatically populate `tag`. 
```
{% if href %}
  {% set tag = 'a' %}
{% endif %}
```

Makes the call a little cleaner: `{{ library.button(text="A link button", href="#") }}`

doc-site/docs/components/buttons.njk
```diff
@@ -7,8 +7,64 @@
...
+  {% endset %}
+  {{ esds_doc.code_example_pair( example=button_default ) }}
+
+  {% set button_default_disabled %}
+    {{ library.button(text="Default Button Disabled", disabled="true") }}
+  {% endset %}
+  {{ esds_doc.code_example_pair( example=button_default_disabled ) }}
+
+  {% set button_default_link %}
+    {{ library.button(text="Default Button Link", tag="a", href="#") }}
```
kevinmpowell: See my note on simplifying this call if you let the presence of `href` trigger `tag="a"`

library/components/button/button.scss
```diff
@@ -1,23 +1,132 @@
...
+  }
+
+  &:focus,
+  &:focus-within {
+    border: $spirit-border-button solid $spirit-interactive-color-focus;
+  }
+}
+
+// Spririt Button - Secondary Gray
+.spirit-button--secondary-gray {
```
kevinmpowell: When do I use `secondary` vs. `secondary-gray`. I'm wondering if there's a more descriptive name that would clarify this?

library/components/button/button.njk
joesnellpdx: @kevinmpowell if 'input' === tag' constructs a new element <input> without a closing </input> tag. 

Example: 
<input type="submit" class="spirit-button" value="Default Button Submit">
vs:
<button class="spirit-button">Default Button</button>

library/components/button/button.scss
joesnellpdx: @jamesmelzer Any reason why the .yaml representation of $spirit-color-blue-30 is not accurate in tokens.scss with the var $spirit-interactive-focus?

library/components/button/button.scss
joesnellpdx: @kevinmpowell @jamesmelzer @jdrf-digitaldesigner 

--tertiary
--gray

--other?

I'd be partial to tertiary to follow the pattern - does not lock into a 'color,' but may be too abstract? If so, will change tokens from spirit-interactive-color-secondary-grey-xxx to spirit-interactive-color-tertiary-text

library/components/button/button.scss
jamesmelzer: I don't. I'll ping Patrick

library/components/button/button.scss
jamesmelzer: Agreed. I didn't have enough info to add them in the initial pass, but once we have a solid review on button, I'll add them. 

Feedback
- .3s feels a little slow for such a tight interaction, to me.

library/components/button/button.scss
jamesmelzer: User error, no doubt. I'll check it in my branch...

jamesmelzer: @joesnellpdx it's #B3CCFF in Sketch, in `$color-blue-30`, and in `$interactive-color-focus`. Where is it showing up as the wrong value?

library/tokens/tokens.yaml
```diff
@@ -20,8 +20,8 @@ color :
     default : &color-red-100 "#F40050"
     light :   &color-red-35 "#FFA6C4"
 elevation-color : # === structure change ===
-  default :  &elevation-color-default  rgba(0,8,26,16)   # based on color-black
-  blue :     &elevation-color-blue     rgba(0,85,255,16) # based on color-blue-default
+  default :  &elevation-color-default  rgba(0,8,26,0.16)   # based on color-black
+  blue :     &elevation-color-blue     rgba(0,85,255,0.16) # based on color-blue-default
```
jamesmelzer: Good catch!

jamesmelzer:
@joesnellpdx On disabled button, you prevent the hover styling from kicking in on default buttons, but --on-dark it's still occurring.

library/components/button/button.scss
joesnellpdx: @jamesmelzer It 'looks' correct, but when I add the variable to a style (i.e. bacgkround-color: $interactive-color-focus) it results as #0055FF on the front end.

Wondering if the following syntax is being ignored due to integer instead of string (30) in tokens.yaml and falling back to 'default':
 30 :      &color-blue-30      "#B3CCFF"

joesnellpdx: @jamesmelzer 

This is what is compiling into tokens.scss:
`$spirit-interactive-color-focus: #0055FF !default;`

jamesmelzer:
Overall feedback on the code (not my area of expertise): looks good; ask Kevin.

Sink page feedback: What I've seen work well elsewhere is showing multiple copies of each variation, one in each state. This aids in visual review and QA, and suggests visual relationships that aren't obvious otherwise (eg, do all the hover states look/feel compatible?) 

This requires a little redundancy in class structure, since it requires `.spirit-thing--hover, .spirit-thing:hover {}` for each state. @kevinmpowell can weigh in. But then the sink can provide easy-to-review rows of variations, with each row (for button) containing: default, hover, focus, active, disabled. 

See "Button States" on https://comet.discoveryeducation.com/components/buttons-sink.html (which, to be fair, has ten million button variations).

library/components/button/button.scss
jamesmelzer: Yeah, you're correct: it's spitting out all the ones with integers. I'm fixing now.

jamesmelzer: I haven't identified the problem. It parses fine in other repos, but is choking here. I'll have to come back to it in the morning.

library/components/button/button.scss
patrickmckowen: @joesnellpdx let's remove `secondary-gray`.  We don't have solid references or guidelines for it yet.

jamesmelzer:
@joesnellpdx Can you take a look at this token PR before I merge it? 
https://github.com/JDRF/spirit/pull/153

joesnellpdx:
@jamesmelzer @kevinmpowell 

For button toggle -- I am not adding any data attributes, aria attributes, or scripts at this time. Is that correct? 

I assume this is meant to mimic the bootstrap toggle:
https://getbootstrap.com/docs/4.0/components/buttons/#toggle-states

patrickmckowen:
@joesnellpdx it's actually more like the Bootstrap checkbox or radio button: https://getbootstrap.com/docs/4.0/components/buttons/#checkbox-and-radio-buttons

The use case for this is the gift amount options on a donation form.

jamesmelzer:
@joesnellpdx in /library/styles/ I'd expect to add...

- accessibility.scss for .spirit-screenreader-only and .spirit-screenreader-hidden
- typography.scss for a basic font reset and the text crop mixin, should we want to include it
- utils.scss for a toolkit, like a border-box mixin or styles for a common overlay layer used by modal, etc
- maybe color.scss for mixins or styles that reveal color beyond specific components (like error red)
- maybe animations.scss for another toolbox of mixins

joesnellpdx:
@patrickmckowen 

If we are constructing a checkbox implementation like this, I suggest adding to forms - not the same as a 'button'

joesnellpdx:
@jamesmelzer @kevinmpowell 

Ready for re-review.

Notes:
Per 6/21/18 scrum - changed borders to inset box-shadow. 
Added utilities scss file for general styles - do not want to cloud button space
Updated button tokens ( removed those fixed in tokens branch )
Per 6/21/18 scrum - removed toggle button, but kept tokens for toggle checkbox in forms
Manually added svg placeholder - until spriting is setup ( will touch base with Kevin on that next week )
A few other minor tweaks

joesnellpdx:
Since new buttons added to sketch - assigning back to self and put task back to in progress.

jamesmelzer:
I'll check the new designs and update the tokens if needed. I don't think we've added any new colors, but the `secondary-gray` has to change and a clear `minimal` is needed.

kevinmpowell:
Responding to Joe's questions posed on the original issue #59 
> Is there a way to prefix svg icons? I added icons, but prefixing with 'ie'. I'm guessing there is a simple update to a gulp task.

I don't understand this question. Are you talking about prefixing filenames?

> How do I attach a commit to an issue ( like Kevin's above), or is this not a big deal?

It is not a big deal, but it makes it easier to find things. Just put a pound sign and the issue number, like `#59` in your commit messages and GitHub will link it up.

> I added a simple doc-site template, but not sure if that is part of this task. Should I continue with that?

I assume you're referring to `doc-site/docs/components/buttons.njk` - that's fine, but typically hold off on working on those pages until we reach the [PUBLISH] step of the process. Typically at the [BUILD] step we just build out the sink pages in `library`. There's a [DOC] step between [BUILD] and [PUBLISH] where we determine what the content and flow of the `doc-site` page will be.

> I added fonts, but not able to accurately test text-crop mixin so line-height still at 1

Go ahead and commit this, I have a separate issue to add the proper values in the next sprint.

> Border for focus is using inset box shadow per our discussion last week - going to come up again with borders on form inputs

This makes sense to me. Do you have any concerns about using box shadow?

> I did not add any font loading enhancements - just @font-face for now ( see: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization )

Again this is fine for now. Can you check the backlog for a task to revisit and optimize? If none can you create one?

> Updated / added some tokens - please review to make sure my assumptions are correct
I'll comment inline on the file

> Nujucks macro - new to this - please review implementation - got complex
I'll add comments inline on the file

> Where do engineering / sketch workflow docks end up? Goal to put into doc-site or are they internal?
That depends on the doc. More often than not, they're valuable to actually publish to the doc site. The main exception I can think of is the release guide for actually deploying the doc site. But git workflow, versioning, etc are all valuable docs for contributors to reference.

> To clarify, all high level conversation on PRs, tickets, issues will be in the associated issue ( here )?
Good question. Clarifying the acceptance criteria before work on the issue begins or the PR is created should go on the issue. After the PR has been created let's migrate future conversation to the PR itself.

kevinmpowell:
Re: toggle button. If this is going to be built I would suggest building this out now. In my experience the toggle button often needs to look identical to the standard button. If they're used side by side in a toolbar for example. 

Since the markup to support `input[type="checkbox']` and `input[type="radio"]` buttons can vary quite a bit from `<a>` or `<button>` based buttons, building it now is actually valuable and will prevent potential breaking changes in a future release.

This might be worth bringing up in Friday's critique.

library/components/button/button.scss
kevinmpowell: I also found a bug. I made the box shadow hot pink so it's easier to see. When I hover the shadow grows, and when I click the shadow shrinks, but when I release but remain hovered I expect the shadow to return to the `:hover` state and it doesn't unless I mouse out of the button and re-enter.
active-release-issue

* [ ] Fix `:active` released hover issue

library/components/button/button.scss
```diff
@@ -1,23 +1,463 @@
...
+
+// Transitions
+$spirit-button-transition-duration: .3s;
+$spirit-button-transition-timing-function: ease-out;
+
+// Spririt Button - Default
 .spirit-button {
-  background: blue;
-  border-radius: 5px;
+  // sass-lint:disable property-sort-order, no-vendor-prefixes
```
kevinmpowell: Is it possible to move this to just above the `-webkit` vendor prefix? that way sort order's still enforced for the rest of the block?

* [ ] Move sass-lint:disable if possible

library/components/button/button.scss
joesnellpdx: @kevinmpowell 

I cannot duplicate. Tested on safari, chrome, ff, and ie 11. Chrome shows focus box-shadow inset, while the others do not ( unless focused via tab ). In all cases, the box shadow goes back to hover state except for chrome - where it remains as :focus ( 0 box shadow ).

Please clarify.

library/components/button/button.njk
```diff
@@ -1,7 +1,85 @@
...
+      {% if icon %}
+        {% if not iconlabel %}
+          {% set iconlabel = icon %}
+        {% endif %}
+        <svg role="img" aria-label="{{iconlabel if iconlabel}}" class="spirit-icon--{{icon if icon }}" xmlns="http://www.w3.org/2000/svg" >
+            <use xlink:href="/icons/spirit.svg#{{icon if icon}}"></use>
+        </svg>
+      {% endif %}
+      {% if loader %}
+        {% if class and 'spirit-button--loading' in class %}
```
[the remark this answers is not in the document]
joesnellpdx: @kevinmpowell 

The classes for success and danger will work out of the box so long as there are no intended hidden / visible text or icons. See the buttons in the 'feedback' button group. These are the 'simple' buttons.

Whether the button content is an icon or text is a macro variable. 

Hidden icons, default text, or feedback text that can be hidden or shown are additional options set by the macros ( controlled by feedbackType and loader variables ). 

For the text elements (when feedbackType and loader variables are used), those are in the dom depending on the variables passed to the macro ( if needed ). What would have to change is the 'aria-hidden' declaration for each one. This cannot be changed via css. 

For loader - because this is an svg animation, it does need to be added to the dom. The only variable required here for the macro is `loader=true` - this adds the dom element (svg). 

Options:
1) Add the loader svg for all buttons regardless if needs a loader or not. 
2) Use pure css instead of svg to create a loading element - can use classes only.
3) Use svg that we can load via css and animate via transition only ( spin ) - can use css classes only

I'm trying to avoid using scripts if at all possible (besides having to change the classname and the 'aria-hidden' declarations ).

Let me know if I'm missing something, need clarification, or not understanding.

See existing loader (inspiration) for reference here:
http://jdrf.design/dist/components.html#01-buttons

joesnellpdx:
@kevinmpowell 

Thanks for answering questions! I believe I have addressed all your issues by my latest commit or by comment response.

Not sure if I should move to Peer Review, however, as there are dependencies we will from the new branches assigned. OR - so I get a peer review from the existing state, get it merged into develop, and add new components (icon macros) when those are available? Fine with either approach.

I assume leave in limbo until I can merge in those other updates.

CC: @jamesmelzer

library/components/button/button.njk
```diff
@@ -1,7 +1,85 @@
...
+  {% if href %}
+    {% set tag = 'a' %}
+  {% endif %}
+  {% if success %}
+    {% set feedbackType = 'success' %}
+  {% elif danger %}
+    {% set feedbackType = 'danger' %}
+  {% endif %}
+  {% if 'input' === tag  %}
+    <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" value="{{icon if icon}}{{text if text}}" {{ 'data-loader=' + loader if loader }} {{ 'data-success=' + success if success }} />
```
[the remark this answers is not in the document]
joesnellpdx: @kevinmpowell Good catch! No you can not. Removing from `input` based buttons.

library/components/button/button.scss
joesnellpdx: @kevinmpowell I tried that to start and it barfs on the sort order even though within ignore declaration. 

Options:
1) leave is is
2) add declaration to a mixin to output in @include and remove sass-lint:disable
3) other

library/components/button/button.njk
```diff
@@ -1,7 +1,85 @@
...
+      feedbackText=false) %}
+  {% if href %}
+    {% set tag = 'a' %}
+  {% endif %}
+  {% if success %}
+    {% set feedbackType = 'success' %}
+  {% elif danger %}
+    {% set feedbackType = 'danger' %}
+  {% endif %}
+  {% if 'input' === tag  %}
```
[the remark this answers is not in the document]
joesnellpdx: @kevinmpowell 

Would love your assistance here.

The issue with input:
1) Cannot have icons in the text
2) Cannot animate icons as pseudo / background elements
3) Does not have closing tag

^^ those are the main concerns with the diff. If those can be resolved, can include in a single output. The main challenge being the loader + feedback text.

PR #158, opened 2018-06-27
Feature/42 space

joesnellpdx:
- add space stack classes with helpers function in scss
- add space stack mixin to output margins with tokens


library/styles/mixins/spacing.scss
```diff
@@ -0,0 +1,14 @@
+/// Spacestack Mixin
+/// @example @include spacestack( 4, left );
+/// @output margin: 0 0 0 32px;
+/// @param {int} $size [2] - Spirit size multiplier
+/// @param {string} $side [left] - Side of margin
+@mixin spacestack($size: 2, $side: bottom) {
```
jamesmelzer: "spacestack" is a misleading name when the space can be on any side.

library/styles/mixins/spacing.scss
```diff
@@ -0,0 +1,14 @@
...
+  $margin: $size * $spirit-space-inset-1-x;
+
+  @if $side == left {
```
jamesmelzer: The helper class use the name "right" but this uses the name "left", yet they provide the same 0 0 0 X space. 

Best guess: left-aligned content is much more frequent, so if I had to pick one, I'd use `inline-left` which pads the right side (0 X 0 0).
Or, support both and throw in a comment to avoid confusion.

library/components/button/button.scss
```diff
@@ -8,6 +8,7 @@ $spirit-button-transition-timing-function: ease-out;
 // Spririt Button - Default
 .spirit-button {
   @include box-sizing(border-box);
```
kevinmpowell: Not related to this pr, but namespace this mixin `spirit-box-sizing`
* [ ] Namespace box sizing mixin

library/styles/spirit.scss
```diff
@@ -1,4 +1,8 @@
-@import '../tokens/tokens';
+@import '../tokens/tokens'; 
+
+@import 'mixins/mixins-control';
```
kevinmpowell: Drop the `mixins-control` and `helpers-control` files. And directly `@import` the additional scss files into this base file. I don't think we need to be overly modular with our `@import`'s right now, and `mixins-control` and `helpers-control` are obscuring relatively straightforward `@import`'s.

* [ ] Remove `mixins-control` and `helpers-control` files and directly `@import` files into `spirit.scss`

library/components/button/button.scss
```diff
@@ -8,6 +8,7 @@ $spirit-button-transition-timing-function: ease-out;
...
+  @include setspace();
```
kevinmpowell: Components shouldn't have any opinions about outside space by default. Remove this from `button.scss`

* [ ] Remove stack space definition from `button.scss`

library/styles/mixins/spacing.scss
```diff
@@ -0,0 +1,17 @@
+/// setspace Mixin
```
kevinmpowell: We don't need this mixin. The stack, inline-left, and inline-right tokens are already defined to correctly set all of these values. We should just use those tokens directly:
`margin: $spirit-space-stack-1-x; // margin: 0 0 8px 0;`
`margin: $spirit-inline-left-1-x: // margin: 0 8px 0 0;`
`margin: $spirit-inline-right-1-x; // margin: 0 0 0 8px`;  

* [ ] Delete `spacing.scss` file - use tokens directly

library/styles/helpers/spacing.scss
```diff
@@ -0,0 +1,47 @@
+
+// Create Space Stack classes using tokens.scss
+
+$spacestack: (
```
kevinmpowell: We already have stack tokens we can leverage for this, and the `tokens.scss` file contains an (undocumented) scss map getter for tokens. Here's a simplified version of this helper class generator:
```
@import '../tokens/tokens';

$spirit-stack-tokens: spirit-token('space', 'stack');
@each $variable, $value in $spirit-stack-tokens {
  .spirit-space--stack-#{$variable} {
    margin: $value;
  }
}

$spirit-inline-left-tokens: spirit-token('space', 'inline-left');
@each $variable, $value in $spirit-inline-left-tokens {
  .spirit-space--inline-left-#{$variable} {
    margin: $value;
  }
}


$spirit-inline-right-tokens: spirit-token('space', 'inline-right');
@each $variable, $value in $spirit-inline-right-tokens {
  .spirit-space--inline-right-#{$variable} {
    margin: $value;
  }
}
```

* [ ] Refactor space helper class generation using built-in token scss map getter.

library/styles/helpers/spacing.scss
```diff
@@ -0,0 +1,47 @@
+
```
kevinmpowell: Since we won't be using these utility classes internally within the library (we'll use the tokens directly instead) let's move this outside of the compilation path for the rest of the library. Let's put this in a peer file to `/styles/spirit.scss` called `/styles/spirit-helper-classes.scss`
That will compile to a separate `spirit-helper-classes.css` file which we can add documentation and usage guidelines around.

* [ ] Move helper class generation to a separate file outside of the `spirit.scss` main library styles

kevinmpowell: Conferred with @patrickmckowen during standup. We agreed to keep this as an `@import` statement in the main `spirit.scss` file for now.

joesnellpdx: @kevinmpowell 

Updates made.

Merging develop into branch and resolving conflicts ( stated below ) results in compiling issues. Issues with compiling should be fixed in develop before resolving merge conflicts here.

joesnellpdx: @kevinmpowell Requested changes addressed. Please confirm.

Issue #163, opened 2018-06-29
Button - fix error with extra quote in span


joesnellpdx:
- fix extra quote in span on line 71

joesnellpdx:
@kevinmpowell @jamesmelzer 

Squashed a bug - may need you to merge if approved as today I am done
SEE: 
https://github.com/JDRF/spirit/pull/164/files

joesnellpdx:
@kevinmpowell 
SEE:
https://github.com/JDRF/spirit/pull/164

joesnellpdx:
@jamesmelzer 

Approved, merged, deleted

PR #164, opened 2018-06-29
#163: remove quote

joesnellpdx:


library/components/button/button.njk
```diff
@@ -68,7 +68,7 @@
             </svg>
           </span>
         {% endif %}
-        <span class="spirit-button__default-text" {{'aria-hidden=true' if 'success' === feedbackType }}">
+        <span class="spirit-button__default-text" {{'aria-hidden=true' if 'success' === feedbackType }}>
```
kevinmpowell: This looks straightforward, but can we flip the if statement so it reads "if feedback type equals success"?
* [ ] Flip if statement so variable comes before test value

joesnellpdx: @kevinmpowell 

Made change, but not showing up in PR:
https://github.com/JDRF/spirit/compare/fix/163-button-fix-error-quote

joesnellpdx:
@kevinmpowell Your change requests have been made. Please confirm.

Issue #54, opened 2017-08-28
Iconography


jamesmelzer:
### Deliver

* Icon macro in nunjucks (treat as component)
* build process (use ESDS)
* Icon sprite with 3-5 sample icons

JodiWarren:
@jamesmelzer @kevinmpowell 
Couple of questions about the icon component:

- Do you want to be able to set the colour/size/anything else at the point of insertion?
- Are there any usage examples which require their own components, or variations of the core component? For example, should I make an icon-badge variation, with the icon inside a solid shape?

jamesmelzer:
@JodiWarren I've generally that handled within the icon itself, to ensure consistent sizing and delivery across encircled/open icons. Let's hold off and see how icons shape up in design.

kevinmpowell:
> Do you want to be able to set the colour/size/anything else at the point of insertion?

No, we should be handling all of that via SCSS.

PR #166, opened 2018-07-05
#54: Iconography

JodiWarren:
- Icon macro & styles
- Bring icons which are inline in text onto the text baseline
- Icon examples

### DIFF
--- a/library/components/icon/icon.njk
+++ b/library/components/icon/icon.njk
@@ -1,0 +1,25 @@
+{% macro icon(
+    icon=false,
+    iconLabel=false,
+    class=false,
+    decorative=false
+    ) %}
+    {% if icon %}
+        {% if not iconLabel %}
+            {% set iconLabel = icon %}
+        {% endif %}
+        <svg
+            aria-label="{{ iconLabel if iconLabel }}"
+            class="spirit-icon{{ ' ' + class if class }}"
+            {% if decorative %}
+                aria-hidden="true"
+            {% endif %}
+        >
+            <use xlink:href="/icons/spirit.svg#{{ icon if icon }}">
+                {% if iconLabel %}
+                    <title>{{ iconLabel }}</title>
+                {% endif %}
+            </use>
+        </svg>
+    {% endif %}
+{% endmacro %}
--- a/library/components/icon/icon.njk
+++ b/library/components/icon/icon.njk
@@ -1,25 +1,39 @@
-{% macro icon(
-    icon=false,
-    iconLabel=false,
-    class=false,
+{#
+use with
+
+{{ library.icon(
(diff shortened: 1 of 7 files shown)


JodiWarren:
@kevinmpowell A first pass at the icon macro. Please let me know if there are more practical applications to test. I couldn't find any in the Sticker Sheet file, but I may have missed some.

library/components/icon/icon.scss
```diff
@@ -0,0 +1,23 @@
...
+  @include box-sizing(border-box); // just in case some extra styling gets applied to the icon.
+  display: inline-block;
+  height: 1em;
+  fill: currentColor;
+  line-height: 1em;
+  vertical-align: middle;
+  width: 1em;
+}
+
+h1,
```
kevinmpowell: Per our critique convo let's drop lines 11-23

* [ ] Remove lines 11-23 from `icon.scss`

library/components/icon/icon.njk
```diff
@@ -0,0 +1,25 @@
+{% macro icon(
+    icon=false,
```
kevinmpowell: Pick an icon to use as a default here. My goal with nunjucks macros is that they'll render _something_ when called without any arguments. This allows for rapid prototyping and quick sanity checks when using the components.

You can also drop the if statement on line 7 by setting a default icon.

* [ ] Set a default icon and remove the `if` statement on lines 7 & 24

JodiWarren:
@kevinmpowell @jamesmelzer 
I've made the requested changes.

I've set the default icon to 'check', which is a tick mark and hopefully unlikely to cause offence.

library/components/icon/icon.njk
```diff
@@ -0,0 +1,39 @@
...
+) }}
+
+param: icon {string} [default: check] The name of the required icon
+param: iconLabel {string} [default: false] An accessible label for the icon. Used if the icon is not decorative. If not set, will use icon name.
+param: decorative {boolean} [default: true] If set to false, will make it visible to screen readers.
+param: class {string} [default: false] Add any required classes in a space-separated list. They will be appended to the default "spirit-icon" class.
+
+ #}
+{% macro icon(
+    icon=check,
```
kevinmpowell: @JodiWarren I'm surprised that `check` isn't a string: `'check'`, I didn't know nunjucks would automatically convert to a string. For clarity, please wrap this in quotes.

* [ ] Wrap default icon name in quotes

JodiWarren: @kevinmpowell Whoops, good catch. That's fixed now.

kevinmpowell:
@jamesmelzer can you review this so it can be merged?

Issue #46, opened 2017-08-28
Typography


jamesmelzer:
### Deliver

* Core typographic hierarchy
* Core typographic mixins (e.g. to support extension by Long Form Text)
* Font family support, including setting up web fonts for the library

JodiWarren:
After doing some reading this morning on the matter, my preferred suggestion is to use Zach Leatherman’s Compromise method, as detailed here: https://www.zachleat.com/web/the-compromise/

It’s a three stage load process:
0 - Unstyled Text
1 - Faux-bold/faux italic text (only the normal/roman weight font loaded)
2 - Fully loaded text.

This is to reduce the visual weight of the text reflow. In reality most text is probably going to be at a normal weight, and it’ll make sure that renders well asap.

I installed glyphanger, and created a subsetted version of the Regular weight Graphik font. I used the [A-Za-z0-9] character ranges. It reduces the woff2 version of Regular weight from 37kb to 12kb. This seems like it would be worth doing for that initial load.

joesnellpdx:
@JodiWarren Try merging develop into your branch now.

JodiWarren:
@kevinmpowell @joesnellpdx @patrickmckowen 
I've removed the font loading system and opened a PR for it: https://github.com/JDRF/spirit/pull/266

PR #169, opened 2018-07-10
#46 Typography

JodiWarren:
I've created a demo page showing the typography as specced in the Sketch file:

- Check out this branch
- `cd library`
- `gulp`
- Navigate to: http://localhost:3000/sink-pages/visual-styles/typography.html - may or may not be localhost:3000, check in the terminal.


patrickmckowen:
@JodiWarren I'm not sure if this is something with BrowserStack Local, but sharing just in case: This is what I see when load the type sink page on IE 11 in Browserstack: https://gmkr.io/s/5b450a39686cfe0848e97d36/0

JodiWarren:
@patrickmckowen Good catch, I'll give it a look over as I'm refactoring the mixins.

JodiWarren:
@kevinmpowell @jamesmelzer 
After yesterday's discussion, I've done some heavy refactoring of this code to make better use of the token hierarchy. I've refactored the token hierarchy too.

The mixins are now substantially more complicated but they should be a lot safer and simpler to use. Once you're happy with them, they can be moved into more appropriate files (Kevin, I would be very grateful for your suggestions here)

@patrickmckowen 
Thanks for that catch yesterday. I found the IE11 issues and have squashed them now.

library/tokens/tokens.yaml
```diff
@@ -189,44 +189,65 @@ font :
...
-    heading-level-4 :              16px
-    heading-level-4-breakpoint-l : 24px
-    heading-level-5 :              14px
-    heading-level-5-breakpoint-l : 18px
-    heading-level-6 :              13px
-    heading-level-6-breakpoint-l : 16px
+    subhead:
+      normal:           16px #use body line height
+      breakpoint-l :    24px
+    display:
```
jamesmelzer: This change makes a lot of sense. I wasn't sure if we were going to have multiple breakpoints, so I kept it flat. The added depth makes it more rigorous (and the depth flattens out to similar SCSS in the end, anyway)

JodiWarren:
@kevinmpowell Apologies - I seem to have slightly messed up a couple of the type files with a dodgy rebase. I've pulled the correct info out of `git reflog`, I just need to tidy up the rebase. I'll try and get it pushed up correctly either today or at worst on Monday. Is that ok?

kevinmpowell:
@JodiWarren I'm not blocked. I've got the font installed locally now so it's not inhibiting my work at this point.

library/styles/base/_typography.scss
```diff
@@ -0,0 +1,219 @@
...
+  font-style: normal;
+  font-weight: normal;
+  letter-spacing: normal;
+  line-height: $spirit-font-line-height-body;
+  text-decoration: none;
+  text-indent: 0;
+  text-shadow: none;
+  text-transform: none;
+
+  .spirit-col--dark & {
```
kevinmpowell: Pull out lines 15-18. We'll handle background context sensitivity on a component by component basis. Putting it in the rest that we'll use everywhere will create too much CSS bloat.

* [ ] Remove `.spirit-col--dark` section from font reset mixin.

library/styles/base/_typography.scss
```diff
@@ -0,0 +1,219 @@
...
+    -webkit-font-smoothing: antialiased; // sass-lint:disable-line no-vendor-prefixes
+    color: $spirit-color-white;
+  }
+}
+
+/// setspace Function
```
kevinmpowell: What does the `//setspace Function` comment do?

JodiWarren: This is part of the sassdoc documentation: http://sassdoc.com/

It's just basically a namespace. If you were to use the sassdoc documentation generation software, it would put it in the right bit.

That said, I just took a look and can't find it on the current site, so it may have been wiped from the latest version of the spec.

JodiWarren:
@kevinmpowell After some mildly hairy gitwork, the branch should hopefully be ok now 😌 That'll teach me to rebase in a hurry.

I've rebased from the current state of develop, which means that I've disabled the buttons import until the reworked variables are integrated into that file.

library/styles/base/_typography.scss
```diff
@@ -0,0 +1,213 @@
...
+) {
+  $line-height-base: ('font', 'line-height');
+
+  $breakpoint-width: spirit-get-breakpoint($breakpoint);
+  $line-height: spirit-get-nested-token($line-height-base, $font-type);
+
+  @include spirit-responsive-value($line-height, 'line-height', $breakpoint, $breakpoint-width);
+}
+
+h1,
```
kevinmpowell: We don't want to style h1-h6 directly with our stylesheets, we want to rely on the mixin internally and provide the utility classes externally.

* [ ] Remove base HTML element styling

library/styles/base/_typography.scss
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/// @group Function
+/// Returns a deeply nested token based on two lists.
+/// @example spirit-get-nested-token(('font', 'size'), ('display', 'level-1;'));
+/// @require {function} spirit-token
+/// @param {list} $parent A comma-separated list of strings which represent tokens.
+/// @param {list | string} $children A single token or comma-separated list of strings which represent tokens.
+/// @return {any} Returns a token result.
+@function spirit-get-nested-token($parent, $children) {
```
kevinmpowell: The `spirit-token` mixin already gets nested tokens. This mixin needs to be renamed to more clearly articulate what it's doing. Is it typography specific?

Upon further digging into the code I believe this can be removed completely. Likely due to your change in the tokens.yaml structure

* [ ] Remove `spirit-get-nested-token` mixin.

library/styles/base/_typography.scss
```diff
@@ -0,0 +1,213 @@
...
+/// @param {string} $breakpoint ['l'] The name of the breakpoint from the token file.
+/// @output Will output result of spirit-responsive-value mixin.
+@mixin spirit-font-size-responsive(
+  $font-type: ('heading', 'level-1'),
+  $breakpoint: 'l'
+) {
+  $font-size-base: ('font', 'size');
+
+  $breakpoint-width: spirit-get-breakpoint($breakpoint);
+  $font-size: spirit-get-nested-token($font-size-base, $font-type);
```
kevinmpowell: Replace this line with:
`$font-size: spirit-token('font', $font-size-base, $font-type);`

* [ ] Remove calls to `spirit-get-nested-token`

library/styles/base/_typography.scss
```diff
@@ -0,0 +1,213 @@
...
+/// @param {string} $breakpoint ['l'] The name of the breakpoint from the token file.
+/// @output Will output result of spirit-responsive-value mixin.
+@mixin spirit-font-line-height-responsive(
+  $font-type: ('heading', 'level-1'),
+  $breakpoint: 'l'
+) {
+  $line-height-base: ('font', 'line-height');
+
+  $breakpoint-width: spirit-get-breakpoint($breakpoint);
+  $line-height: spirit-get-nested-token($line-height-base, $font-type);
```
kevinmpowell: Replace this line with:
`$line-height: spirit-token('line-height', $line-height-base, $font-type);`

* [ ] Remove calls to `spirit-get-nested-token`

library/styles/base/_typography.scss
JodiWarren: @kevinmpowell This mixin was necessary because `spirit-token` accepts and operates over map, whereas we're providing a list here. They look similar but are different data types in SassScript. In theory, I could've done some of this work in each of the font-size and line-height mixins, but I spun this out to keep it a bit DRYer.

JodiWarren:
@kevinmpowell This refactor as a whole was instructed from comments on a call a couple of weeks ago where I was asked to find a solution which didn't involve the long variable names. I'm **super** happy to roll back to a simpler solution. Simpler, less fragile, less clever code is always going to be a better solution.

In my opinion, combining media queries shouldn't be a result of the developer, but instead a tooling job. One of the highlights of Sass is nested mixins. If you agree, there's an excellent PostCSS plugin that I've used in the past: https://github.com/hail2u/node-css-mqpacker

JodiWarren:
@kevinmpowell I've created some reusable mixins for the type styles. While this may seem initially redundant, it will make it much easier to reuse those styles in other components, like in the long form text section. This is in preference to using the clunky and error-prone `@extends` command.

If you're happy with this, can you please prioritise approving/merging it? It will make future development of type-related components substantially easier.

library/styles/mixins/typography.scss
```diff
@@ -23,3 +23,128 @@
   text-shadow: $text-shadow;
   text-transform: $text-transform;
 }
+
+
+@mixin spirit-display-1 {
+  @include spirit-typography-reset;
```
kevinmpowell: Pass these properties into the typography reset so that we're not setting properties like `color`, `font-size`, etc. more than once.

Use this syntax:
```
@include spirit-typography-reset(
  $color: $spirit-color-black,
  $font-size: $spirit-font-size-display-level-1-normal,
  $font-weight: $spirit-font-weight-bold
)
```

* [ ] Pass overrides into typography reset calls

library/styles/mixins/typography.scss
```diff
@@ -23,3 +23,128 @@
...
+    font-size: $spirit-font-size-display-level-1-breakpoint-l;
+    line-height: $spirit-font-line-height-display-level-1-breakpoint-l;
+  }
+}
+
+@mixin spirit-display-2 {
+  @include spirit-typography-reset;
+  color: $spirit-color-black;
+  font-size: $spirit-font-size-display-level-2-normal;
+  font-weight: 700;
```
kevinmpowell: * [ ] Replace all instances of `700` with `$spirit-font-weight-bold`

library/templates/sink.njk
```diff
@@ -136,6 +136,47 @@
             font-size: 16px;
         }
     </style>
+    <script>
```
kevinmpowell: Let's move this to a `/scripts` directory in the root of `/library`
Probably as `typography.js`

We'll need this font loader on the `/doc-site` so we'll want to ship this.
The build tooling will concatenate scripts in `/library/scripts` and merge them into a file accessible from the webroot as `/scripts/spirit.js` so you'll be able to replace this block with:
`<script src="/scripts/spirit.js"></script>` 

* [ ] Move font loader script out of template and into `/library/scripts/typography.js`

JodiWarren: @kevinmpowell Because I'm not certain that `fonts` will always be in the same place relative to the scripts folder, I'm going to move `critical-foft-preload-fallback-optional.js` into the `scripts` folder too. It's because it gets referenced by the type loader JS:
 https://github.com/JDRF/spirit/pull/169/files/570467f3bb3a9eb3f5b7ffbc3685419e78399375#diff-6cc84a5c53dea68dee01e55a099398e3R167

PR #172, opened 2018-07-12
Feature/64  form components

kevinmpowell:
## IMPORTANT
After checking out, run `npm install`. This branch relies on an updated version of `esds-build`.

Sorry about all the icon updates. I fixed a bug in the build tooling that was over-optimizing the source SVG's. 

The forms sink page is at: `/sink-pages/components/forms.html`


jamesmelzer:
I glanced through. Very nice. I'll go over Sketch symbol changes to align with this tomorrow.

library/components/form/form.scss
```diff
@@ -0,0 +1,908 @@
...
+// Field Group
+$spirit-space-field-group-margin:             $spirit-form-group-stack;
+$spirit-space-field-group-horizontal-margin:  0 16px 0 0;
+
+.spirit-form__field-group {
+  @include spirit-typography-reset;
+  margin: $spirit-space-field-group-margin;
+}
+
+.spirit-form__field-group--horizontal {
```
joesnellpdx: @kevinmpowell 

This needs to be enabled via breakpoint ( ie. `horizontal-md` ). Most cases will start as full width, then become gridified.

* [ ]  Create breakpoints for horizontal grouping

library/components/form/form.scss
```diff
@@ -0,0 +1,908 @@
...
+  align-items: flex-start;
+  display: flex;
+
+  & > * {
```
joesnellpdx: @kevinmpowell 

Horizontal grouping children can become janked if the label does not fit ( inputs not aligned ).

screen shot 2018-07-19 at 4 09 03 pm

To resolve:
1. Remove `align-itms: flex-start;` from group--horizontal.
2. Add `display: flex;` and `flex-flow: column nowrap;` to children.
3. Add `margin-top: auto;` to input inside child.

* [ ] Fix misaligned inputs when label extends beyond width of 'grid'

library/docs/sink-pages/components/forms.njk
```diff
@@ -0,0 +1,779 @@
...
+        {% call library.form_fieldset() %}
+            {{ library.form_legend(text="Shipping Address") }}
+            
+            {% call library.form_field_group() %}
+                {{ library.form_checkbox(label="Same as billing address", id='checkout-example-checkbox', checked=true) }}
+            {% endcall %}
+
+            {% call library.form_field_group() %}
+                {{ library.form_label(text="Street", for="full-form-shipping-street") }}
+                {{ library.form_input(disabled=true, id="full-form-shipping-street") }}
```
joesnellpdx: @kevinmpowell 

Is lighter grey enough of an indicator that this input is disabled ( especially without a text )? The addition of the red (`*`) sends the opposite message. Perhaps no background color and a light border. Hover indicator not enough as not all will be using a mouse.

I did not realize was an issue until I tried to click into it. ;-)

Something to discuss with team: CC @jamesmelzer @ismaeltoure7 @patrickmckowen

patrickmckowen: I wonder if the black input labels are sending the wrong message; perhaps changing the label color to Grey will help.

joesnellpdx: @patrickmckowen 

Lightening the black inputs would definitely help. I'm also concerned that the 'difference' between the grays is very subtle. 

This is a different design, but I feel there is more clarity on what is disabled:
https://developer.microsoft.com/en-us/fabric#/components/textfield

That being said - not much difference in Material ( link below ) - but it does require there being a label and that it is more washed out.: 
https://material.io/design/components/text-fields.html#filled-text-field

jamesmelzer: Agreed: lightening the labels one step helps. 
screen shot 2018-07-23 at 8 38 31 am

joesnellpdx: I'm still not sure why disabled has `*`

library/tokens/tokens.yaml
```diff
@@ -14,11 +14,11 @@ color :
     30 :      &color-blue-30      "#B3CCFF"
     default : &color-blue-default "#0055FF"
   green :
-    default : &color-green-67 "#37A941"
+    67 :      &color-green-67 "#37A941"
     light :   &color-green-20 "#BBE9BF"
```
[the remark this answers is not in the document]
kevinmpowell: I believe @jamesmelzer has already adjusted this. I've synced with `develop` and all palette colors are numeric now.

library/docs/sink-pages/components/forms.njk
```diff
@@ -0,0 +1,779 @@
...
+
+    {% call library.form_field_group() %}
+      {{ library.form_label(text="Disabled", required=true) }}
+      {{ library.form_input(placeholder="First name", disabled=true) }}
+      {{ library.form_microcopy(text='<sup>*</sup>Adds &ldquo;Not Allowed&rdquo; cursor to disabled state.') }}
+    {% endcall %}
+
+    <h5 class="hostile-sink-reset">Hidden Label</h5>    
+    {% call library.form_field_group() %}
+      {{ library.form_input(placeholder="Default") }}
```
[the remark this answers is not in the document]
kevinmpowell: I prefer to author macros at the smallest level of functionality. This gives us flexibility in composition. In any markup examples we author in the docs we will always include `label` and the correct `aria` attributes to make the form accessible.

library/docs/sink-pages/components/forms.njk
```diff
@@ -0,0 +1,779 @@
...
+      {{ library.form_label(text="Error With Micro Copy", required=true) }}
+      {{ library.form_input(placeholder="First name", valid=false) }}
+      {{ library.form_field_error(text="Helpful error message.") }}
+      {{ library.form_microcopy(text="Micro copy goes here.") }}
+    {% endcall %}
+
+    {% call library.form_field_group() %}
+      {{ library.form_label(text="Disabled", required=true) }}
+      {{ library.form_input(placeholder="First name", disabled=true) }}
+      {{ library.form_microcopy(text='<sup>*</sup>Adds &ldquo;Not Allowed&rdquo; cursor to disabled state.') }}
```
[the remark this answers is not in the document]
kevinmpowell: I _could_ see a disabled form field needing to be required if another required field had to be completed to enable the form. I understand that it's not a great best practice for form design, and I wouldn't put an example like this on the actual doc page.

library/components/form/form.njk
```diff
@@ -0,0 +1,455 @@
...
+                {% endif %}
+            </legend>
+        {% endif %}
+        {%- if caller -%}
+            {{ caller() }}
+        {%- endif -%}
+    </fieldset>
+{% endmacro %}
+
+{% macro form_radio_group(
```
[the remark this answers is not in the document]
kevinmpowell: We can't require it within the macro. The docs will cover this as in: http://designsystem.morningstar.com/components/forms.html#radio-group

I'll fix the sink page examples that are broken.

library/docs/sink-pages/components/forms.njk
kevinmpowell: @joesnellpdx I imagine in this specific case those fields could be prepopulated from the Billing Address entries, but this is likely just an oversight on my part. By default the label macro includes the required asterisk since the majority of the examples have the required asterisk. I will remove.

library/components/form/form.scss
kevinmpowell: It's my understanding from our previous critique that all horizontal form layout will be handled by the grid. @ismaeltoure7 is that your understanding as well? If so, should we remove these examples from the sink page?

joesnellpdx:
@kevinmpowell Per our meeting today and your updates -- approved. Cool stuff!

Issue #161, opened 2018-06-29
Button: Follow Up Pass


jamesmelzer:
Follow up to the first pass at button in #59 

Two tasks:

1. Resolve feedback from https://github.com/JDRF/spirit/pull/152 ( review comments here )
2. Update minimal button based on recent design changes. Note: token values may accomplish some of this. See #160.
3. Add toggle button ( input checkbox / radio ) - see forms in sketch. Addressed in feedback from pull above, but may need to discuss with Kevin. Add to macro and possibly improve macro conditionals.

joesnellpdx:
@kevinmpowell @jamesmelzer 

Per discussion with Kevin today, do we want to remove / alter use of box shadow for border in this ticket? What is technical strategy?

joesnellpdx:
PR again ready for review:
https://github.com/JDRF/spirit/pull/173#pullrequestreview-138656953

- [ ] @patrickmckowen 
- [ ] @jamesmelzer 
- [ ] @kevinmpowell

joesnellpdx:
Added @patrickmckowen and @james for PR:
https://github.com/JDRF/spirit/pull/173#pullrequestreview-139939683

PR #173, opened 2018-07-17
#161: Button Follow Up Pass

joesnellpdx:
**Issues from previous task:**

1. Letter spacing token - left in button.scss for now - should this be added to tokens?
2. Released hover issue - cannot replicate - but do see a difference between Chrome and other browsers. Only chrome showing 'focus' state on mouse click ( but not released hover issue ).
3. Add toggle buttons ( checkbox & radio )
4. Refactor so `input` buttons and `a` and button buttons all support the same functionality see https://github.com/JDRF/spirit/pull/152 for questions ( last comment )

### DIFF
--- a/library/components/button/button.njk
+++ b/library/components/button/button.njk
@@ -1,33 +1,32 @@
 {% macro button(
+      id=false,
       class=false,
       disabled=false,
       text=false,
       icon=false,
       iconLabel=false,
-      tag="button",
+      el="button",
       type=false,
       href=false,
       loader=false,
       feedbackType=false,
       feedbackText=false,
-      inputName=false,
-      inputId=false,
-      inputValue=false
+      name=false,
+      value=false
       ) %}
   {% if href %}
-    {% set tag = 'a' %}
+    {% set el = 'a' %}
   {% endif %}
-  {% if tag === 'input'  %}
-    {% if (type === 'checkbox') or (type === 'radio') %}
-      <{{tag}} class="spirit-button-input-toggle" {{ 'type=' + type if type }} {{ 'name=' + inputName if inputName }} {{ 'id=' + inputId if inputId }} {{ 'value=' + inputValue if inputValue }} tabindex="-1"/>
-      <label {{ 'for=' + inputId if inputId }} class="spirit-button spirit-button--toggle{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" tabindex="0">
+  {% if ( el === 'checkbox' ) or ( el === 'radio' ) %}
+      <input class="spirit-button-input-toggle" {{ 'type=' + el }} {{ 'name=' + name if name }} {{ 'id=' + id if id }} {{ 'value=' + value if value }} tabindex="-1"/>
+      <label {{ 'for=' + id if id }} class="spirit-button spirit-button--toggle{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" tabindex="0">
         {{ text if text }}
       </label>
-    {% else %}
-      <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" value="{{text if text}}" {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} />
-    {% endif %}
-  {% else %}
-    <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" {{ 'href=' + href if href }} {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} {{ 'disabled' if disabled }}>
+  {% elif ( el === 'submit' or el === 'reset' ) %}
+    <input {{ 'id=' + id if id }} {{ 'type=' + el }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" value="{{text if text}}" {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} />
+  {% elif el %}
+    {# Using type variable for <button type="submit"> or <a type="button"> options #}
+    <{{el}} {{ 'id=' + id if id }} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" {{ 'href=' + href if href }} {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} {{ 'disabled' if disabled }}>
       {% if icon %}
         {% if not iconLabel %}
           {% set iconLabel = icon %}
@@ -79,12 +78,12 @@
             </svg>
           </span>
         {% endif %}
-        <span class="spirit-button__default-text" {{'aria-hidden=true' if 'success' === feedbackType }}">
+        <span class="spirit-button__default-text" {{'aria-hidden=true' if 'success' === feedbackType }}>
           {{ text if text }}
         </span>
       {% else %}
         {{ text if text }}  
       {% endif %}
-      </{{tag}}>
-    {% endif %}
+      </{{el}}>
+  {% endif %}
 {% endmacro %}
--- a/library/components/button/button.njk
+++ b/library/components/button/button.njk
@@ -10,16 +10,17 @@
       loader=false,
       feedbackType=false,
       feedbackText=false,
-      inputname=false,
-      inputid=false
+      inputName=false,
+      inputId=false,
+      inputValue=false
       ) %}
   {% if href %}
     {% set tag = 'a' %}
   {% endif %}
   {% if tag === 'input'  %}
     {% if (type === 'checkbox') or (type === 'radio') %}
-      <{{tag}} class="spirit-button-input-toggle" {{ 'type=' + type if type }} {{ 'name=' + inputname if inputname }} {{ 'id=' + inputid if inputid }} />
-      <label {{ 'for=' + inputid if inputid }} class="spirit-button spirit-button--label{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}">
+      <{{tag}} class="spirit-button-input-toggle" {{ 'type=' + type if type }} {{ 'name=' + inputName if inputName }} {{ 'id=' + inputId if inputId }} {{ 'value=' + inputValue if inputValue }} tabindex="-1"/>
+      <label {{ 'for=' + inputId if inputId }} class="spirit-button spirit-button--toggle{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" tabindex="0">
         {{ text if text }}
       </label>
     {% else %}
--- a/library/components/button/button.njk
+++ b/library/components/button/button.njk
@@ -9,12 +9,22 @@
       href=false,
       loader=false,
       feedbackType=false,
-      feedbackText=false) %}
+      feedbackText=false,
+      inputname=false,
+      inputid=false
+      ) %}
   {% if href %}
     {% set tag = 'a' %}
   {% endif %}
   {% if tag === 'input'  %}
-    <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" value="{{text if text}}" {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} />
+    {% if (type === 'checkbox') or (type === 'radio') %}
+      <{{tag}} class="spirit-button-input-toggle" {{ 'type=' + type if type }} {{ 'name=' + inputname if inputname }} {{ 'id=' + inputid if inputid }} />
+      <label {{ 'for=' + inputid if inputid }} class="spirit-button spirit-button--label{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}">
+        {{ text if text }}
+      </label>
+    {% else %}
+      <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" value="{{text if text}}" {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} />
+    {% endif %}
   {% else %}
     <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" {{ 'href=' + href if href }} {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} {{ 'disabled' if disabled }}>
       {% if icon %}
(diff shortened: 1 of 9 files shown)


jamesmelzer:
Joe's Q: Letter spacing token - left in button.scss for now - should this be added to tokens?
James' A: This seems button-specific for now. Let's see if it generalizes to be applied elsewhere before putting in tokens.

library/components/button/button.njk
```diff
@@ -9,12 +9,23 @@
...
+      inputName=false,
+      inputId=false,
+      inputValue=false
+      ) %}
   {% if href %}
     {% set tag = 'a' %}
   {% endif %}
   {% if tag === 'input'  %}
-    <{{tag}} {{ 'type=' + type if type }} class="spirit-button{{ ' ' + class if class }}{{ ' spirit-button--is-icon' if icon and not text}}" value="{{text if text}}" {{ 'data-loader=' + loader if loader }} {{ 'data-type=' + feedbackType if feedbackType }} />
+    {% if (type === 'checkbox') or (type === 'radio') %}
```
kevinmpowell: I understand the need for this branching structure so your next sibling selectors work, but I'd like to see the top-level element remain `.spirit-button`. The way I've achieved this in the past is to wrap the `input` and `label` inside a `div.spirit-button`.

So that regardless of which element I'm using for a button the outermost wrapper is always the block level element in BEM. I leave it up to you to decide the LOE to refactor what you've got. 

Here's a reduced test case to show what I mean: https://codepen.io/kevinmpowell/pen/WKGOXd?editors=1100

* [ ] Consider refactoring `radio` and `checkbox` button markup

library/docs/sink-pages/components/buttons.njk
```diff
@@ -123,17 +123,68 @@
...
+                    ## Toggle Disabled
+                {% endfilter %}
+                <div class="spirit-button-group-toggle">
+                    {{ library.button(class="spirit-button--disabled", text="$10", tag="input", type="checkbox", inputName="checkbox", inputId="option1", inputValue="10") }}
+                </div>
+
+                {% filter markdown(true, "hostile-sink-reset") %}
+                    ## Toggle Group
+                {% endfilter %}
+                <div class="spirit-button-group-toggle">
```
kevinmpowell: I'd expected `spirit-button-group` to be its own component with its own macro. The way it's currently build the `spirit-button-group-toggle` class seems to be serving two purposes:
1. Wrapping the input and label elements for radio/checkbox buttons
2. Adding additional "group" visuals to buttons (borders, etc.)

I expect it to do #2 but not #1. See my other comment about refactoring the `radio` and `checkbox` button markup.

I'd also make a new component called "Button Group" since it often warrants its own documentation in the doc site. Give it its own sink, macro, and scss file.

You can use the `caller` syntax in nunjucks to create this type of "wrapper" component.

```
{% macro button_group() %}
  <div class="spirit-button-group">
    {{ caller() }}
  </div>
{% endmacro %}
```
And then to use it:
```
{% call library.spirit_button_group() %}
  {{ library.button() }}
  {{ library.button() }}
  {{ library.button() }}
{% end call %}
```

* [ ] Make button group a separate component

library/tokens/tokens.yaml
```diff
@@ -37,12 +37,21 @@ icon-color :
...
   primary : *color-blue-100
   on-dark : *color-white
+  secondary : *color-grey-95
   text : 
     primary : *color-blue-100
     on-dark : *color-white
   disabled :
     foreground : *color-grey-70
     background : *color-grey-98
+  hover :
```
kevinmpowell: I'm not convinced these tokens belong here. Are they going to be used outside of button anywhere?
If so, let's come up with some more descriptive names. I don't know what `hover-minimal-background` means or when I would use it.

* [ ] Move tokens back into button.scss file or rename them to be more broadly applicable

library/components/button/button.scss
```diff
@@ -103,6 +135,11 @@ $spirit-button-transition-timing-function: ease-out;
       color: $spirit-interactive-color-disabled-foreground;
     }
   }
+
+  .spirit-container--dark .spirit-button-input-toggle:checked +&,
+  .spirit-col--dark .spirit-button-input-toggle:checked +& {
+    box-shadow: $spirit-elevation-1 $spirit-elevation-color-blue, inset 0 0 0 $spirit-border-button $spirit-interactive-color-focus-soft;
```
kevinmpowell: I would like to see the border return to being controlled by the `border` property and reserve the `box-shadow` for only the `:focus` state. 

This will require design approval from @ismaeltoure7 as it may affect the button height, but it makes the CSS much easier to maintain and reason about.

I've also had issues with `box-shadow` based borders breaking when the browser content is zoomed.

* [ ] Refactor button borders to use the `border` property instead of `box-shadow`

library/components/button/button.scss
```diff
@@ -252,46 +290,47 @@ $spirit-button-transition-timing-function: ease-out;
...
     box-shadow: $spirit-elevation-1 transparent, inset 0 0 0 $spirit-border-button $spirit-interactive-color-focus-soft;
   }
 
   &:active,
   &.spirit-button--active {
-    box-shadow: $spirit-elevation-2 $spirit-elevation-color-default, inset 0 0 0 $spirit-border-button-secondary $spirit-color-grey-95;
+    box-shadow: $spirit-elevation-2 $spirit-elevation-color-default, inset 0 0 0 $spirit-border-button-secondary $spirit-interactive-color-secondary;
   }
 
   &:disabled {
```
jamesmelzer: The disabled button still moves when hovered. 
&:disabled should include an override for this (or it should include a ":not()" in the original statement):
`&:hover, &--hover {transform: translateY(-1px);}`

joesnellpdx:
Note to self:
Before assigning back to review--
1. Assign Patrick to review
2. Show options for border vs box-shadow

library/components/button/button.njk
```diff
@@ -9,12 +9,23 @@
...
-      feedbackText=false) %}
+      feedbackText=false,
+      inputName=false,
+      inputId=false,
+      inputValue=false
+      ) %}
   {% if href %}
     {% set tag = 'a' %}
   {% endif %}
   {% if tag === 'input'  %}
```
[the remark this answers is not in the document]
joesnellpdx: @kevinmpowell This is complete - but left 'type' for instances like <a type="button"> and <button type="submit"> - left comment in code as well.

library/components/button/button.scss
joesnellpdx: @kevinmpowell @jamesmelzer 

Did not refactor, but added a demo of the standard / primary button in the second column.

joesnellpdx:
@jamesmelzer 
@patrickmckowen 
@kevinmpowell 

This is ready for review. 

James - cannot 'check' your request for the on-dark bg, but that has been updated

Kevin - your requests have been updated with a comment above regarding 'type'.

See the second column for the demo using border instead of box-shadow as requested - buttons are the wrong size here. To fix, could add tokes / custom styles for padding ( padding - border ), or other potential options.

Let's discuss.

jamesmelzer:
@patrickmckowen 
That `text-shadow` value (0.08) seems very low. I can't see it. Consider darkening? 
Adding `transform: translateY(1px);` to active looks great.

patrickmckowen:
@jamesmelzer the purpose of the text-shadow is to soften the vibrating effect white has on that blue, not to create an elevated style.  With that in mind, do you still recommend darkening?

jamesmelzer:
@patrickmckowen No, that makes sense. Ignore my suggestion.

kevinmpowell:
@joesnellpdx re keeping `type` in the API. I think you could narrow your API by omitting `type` and simply allowing `el` to receive `'submit'` and `'reset'` as possible values. Internally in the component you could then programmatically set the `type` attribute without exposing it to macro users. Not a huge deal, but change it if you want.

joesnellpdx:
@patrickmckowen @kevinmpowell @jamesmelzer 

1. Antialising - added to button, but should this be in the `spirit-typography-reset` mixin @kevinmpowell ? What is the intention of this mixin - use everywhere or establish a global spirit reset somewhere?

2. Letter spacing. There are several 'custom' variables on the buttons scss file. I have added the `$spirit-text-shadow-small: 1px 2px rgba(0, 8, 26, .08);` variable here as well. Do we want to add to the global tokens file @jamesmelzer ? How about the naming? 

3. Chrome focus. I don't think we can remove the border (whether it is `border` or `box-shadow`) on focus on chrome ( or other browsers not on MAC ) with simple css. It is a browser specific behavior. We would have to only apply a focus class on keyboard focus via custom javascript - which is not recommended. @kevinmpowell, unless you have seen something different? See: https://gist.github.com/cvrebert/68659d0333a578d75372

Other than the these items, I believe this branch is done.

Please respond to comments above.

library/components/button/button.scss
```diff
@@ -1,12 +1,29 @@
 // Letter Spacing
-$spirit-button-font-letter-spacing-15: .15em;
+$spirit-button-font-letter-spacing-125: .125em;
```
patrickmckowen: @joesnellpdx two things here: 

1.  The computed value should be .2px not 2px
2. This is value is not relative and shouldn't scale up i.e. if someone zoomed 200% the spacing between the letters shouldn't increase proportionally.  With that in mind, do you still suggest we use a relative unit instead of pixels?

library/components/button/button.scss
```diff
@@ -49,18 +67,22 @@ $spirit-button-transition-timing-function: ease-out;
   &--focus {
     box-shadow: $spirit-elevation-1 $spirit-elevation-color-blue, inset 0 0 0 $spirit-border-button $spirit-interactive-color-focus-soft;
     outline: 0;
+    z-index: 1;
   }
 
   &:active,
   &--active {
     box-shadow: $spirit-elevation-2 $spirit-elevation-color-blue, inset 0 0 0 $spirit-border-button $spirit-interactive-color-primary;
+    transform: translateY(-1px);
```
patrickmckowen: @joesnellpdx this should be 1px not -1px.  The intention here is to give the illusion that the button is being pressed down on click.

patrickmckowen:
@joesnellpdx regarding the focus ring with Chrome on click:  I quickly checked and Airbnb, Stripe, Shopify, and Apple all have a distinct focus state on buttons when tabbing but do not show on click.  I'd like to learn more about why removing it on chrome is not recommended.  Thanks!

jamesmelzer:
* Shopify/Polaris is defining the active state to override the focus state (looks like using CSS source order), so clicking hides any visible indicator that Chrome is also focusing. 
* Whatever AirBnB's guidelines say, all their buttons get a visible focus ring on click (in Chrome). 

I don't know if it matters that these are both React-based applications

patrickmckowen:
@jamesmelzer the Invite Friends button on this page does not get a visible ring on click in chrome, for me at least. https://www.airbnb.com/trips

Bottom line: the focus ring is not acceptable for buttons on click.  Happy to continue discussing options for how to resolve this.

jamesmelzer:
We should be able to force the active+focus state's styling. Irrelevant on other browsers, but will do what we want in Chrome. 
Thoughts @joesnellpdx ?

library/components/button/button.scss
joesnellpdx: @patrickmckowen 

Do be clear, if the font size were 160px, you wouldn't want the letter spacing to be 2px - you would still want .2px? Otherwise, I'd use relative sizes as em is relative to the font size.

patrickmckowen: @joesnellpdx exactly, the .2px letter spacing is only to help with legibility at small sizes.  If the font was 160px it wouldn't need additional letter spacing to be legible e.g. https://gmkr.io/s/5b57aaaeea48c120b7920bc3/0

joesnellpdx:
@patrickmckowen @jamesmelzer @kevinmpowell 

Just tested several of the sites listed - in the cases I tested, they are using javascript to remove focus when a mouseclick is used OR adding focus when the tab key is depressed. 

They are using different scripts fo this. 

This is a **SUPER** common discussion between design, engineering and a11y. 

The a11y argument: someone may be using a mouse that needs to see focus AND you should't rely on javascript for accesibility.

The designer argument: it is not a clean UX and does not fit our design patterns PLUS we feel that with the scripts we can solve the a11y question for most cases.

Here are the options ( as I see them ):
1. Leave as is
2. Add a script that needs to be 'consumed' when using the design system to negate keyboard clicking ( with styles in the design system to handle these script updates ).
3. Do number 2, but have it be a 'requested' step on the doc-site for users of Spirit to add to their own implementations ( recommended script option ). If the script is not implemented, focus states will be visible on click.
4. Use `:focus-visible` - meant to solve this issue, but only available in Firefox - so when (if) chrome and other browsers adopt this pseudo class, we can control styling: https://caniuse.com/#search=focus-visible
5. Remove focus states altogether ( not recommended )

:active + :focus is not reliable as it requires an 'active' state and not all els that will receive focus also will receive an active state. This will likely not solve our challenge. 

Script options:
https://www.npmjs.com/package/focus-visible
I've rolled my own, but the above uses the new focus-visible syntax so I'd recommend that

Happy to investigate, test, discuss further.

patrickmckowen:
@joesnellpdx thanks for this.  I prefer option #2.  I think that with the prevalence of buttons + the popularity of chrome with our user base that it's worth the extra dependency.

kevinmpowell:
@patrickmckowen I disagree with removing the focus behavior on Chrome for a few reasons.
1. Chrome's behavior (in my testing) only applies to actual `<button>` elements. Not `<a>`'s that _look_ like buttons or `<input type='checkbox'>` elements that look like buttons. I think the biggest use case you'll have for an actual `<button>` is the submit button at the bottom of a form where either one of two things will occur:
  * You'll be taken to the next page, meaning you'll see the focus ring for a brief moment before the page reloads
  * Your form will have an error and the validator will shift focus to the first form element that has an error, meaning you probably won't see that focus ring occur at all
Bottom line, it's likely an "edge of an edge" case.
2. The technical solution adds Javascript when we prefer to keep the system JS-free to encourage adoption by product teams.
3. This solution overrides what Chrome has deemed the default behavior for their users. True, there are other sites that are overriding this behavior, but I would estimate that _most_ sites do not, and therefore Chrome users are used to this behavior. In fact, some Chrome users may think a site is broken or a button unresponsive if it doesn't exhibit the same behavior they see on the majority of sites.

I want to be cautious about going down the road of "pixel perfect in every browser"

kevinmpowell:
I stand corrected on my first point. This _does_ apply to `<a>` buttons as well, but if you click those, the page is gonna reload anyway.

patrickmckowen:
@joesnellpdx @kevinmpowell I didn't realize this only applied to `<button>`.

It's not true that people won't see it on form submit; we use loading buttons _because_ some of our forms are slow.  I checked how this looks on the loading state buttons on the sink page and I think it's fine.  

thanks for the explanation.

jamesmelzer:
Weighing in, Kevin's first point is the one I'd like to focus on: buttons generally submit to another page, so in real-world application button's focus state will not be visible to the end user, who will already be on their way to the next page. 

I don't want the team to spin a lot on an issue that may have very little real-world impact. Are we sure anyone will see the state we're discussing? 

I'm also keen to avoid:
- Using JS to achieve pixel-perfect alignment across browsers. 
- A mandatory JS dependency in our "JS is only a placeholder to model behavior" library. 

We can mitigate with CSS - suppress the focus styling when the button is active. And we can mitigate with guidelines - when a button is clicked and on-page validate fires an error alert, move focus to that alert. Neither one requires button JS.

jamesmelzer:
I'm a moment too late in my comment. Disregard.

patrickmckowen:
@joesnellpdx just one small change and I'm all good on buttons!  For the secondary buttons on light background, please update the elevation color to blue.

PR #174, opened 2018-07-20
Feature/111 card

kevinmpowell:
I've added both a cards & a card-grids sink with this update.


library/components/card_grid/card_grid.js
```diff
@@ -0,0 +1,79 @@
+class SpiritCardGrid {
```
joesnellpdx: @kevinmpowell 

Do we want these scripts running regardless of breakpoint? Do we want to activate / destroy depending?

If so, could set css based on modifier class to generate carousel styles (as well as any other a11y helpers like aria-hidden, aria-controls, etc).

library/components/card/card.njk
```diff
@@ -0,0 +1,38 @@
...
+    {% set size_class = "spirit-card--" + size %}
+  {% endif %}
+
+  <a href="{{ href }}" class="spirit-card{{ ' ' + class if class }}{{ ' ' + size_class if size_class }}">
+    {% if img_src %}
+      <span class="spirit-card__image-wrap">
+        {% if img_src_list_item %}
+          <img class="spirit-card__image-list-item" src="{{ img_src_list_item }}" alt="{{ img_alt_text }}">
+        {% endif %}
+        <img class="spirit-card__image" src="{{ img_src }}" alt="{{ img_alt_text }}">
```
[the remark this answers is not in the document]
kevinmpowell: I'm not concerned about crafting the macro for consumption outside the core team. I'll raise this in the next crit to make sure we're all on the same page about it.

library/components/card_grid/card_grid.js
kevinmpowell: I typically avoid wading too deeply into Javascript land, especially for a 1.0. Given the various build tools and frameworks out there many teams already have a breakpoint/js loading solution in their application. My intent with JS (which I hesitate to author in Design Systems) is to merely _demonstrate_ the functionality that is needed to make a component work. We need to have a larger conversation about this in the next critique so we can align on a strategy and package and document the scripts correctly.

kevinmpowell:
> a11y - the carousel. Is it WCAG 2.1 a11y? Should we be concerned about this now ( perhaps a p2 item)? Also, should we be running the scripts on all breakpoints ( even if carousels do not exist )?

@joesnellpdx I'll add some accessibility hooks to the JS demo script I wrote.

joesnellpdx:
@kevinmpowell I see the thumbs up, but not the updates. LMK when updates ready for re-review. Most of comments resolved per discussion outside of ticket.

ismaeltoure7:
@kevinmpowell Looks great! I would reduce the transform: scale to 1.1

library/components/card_grid/card_grid.js
patrickmckowen: I defer to you guys on the specifics of this issue but I will share a couple things:

Cards should _just work_ without requiring the implementing team to have their own JS solutions. 

This is because JDRF relies primarily on external vendors, rather than internal product teams. With internal teams, getting people to adopt the system is a challenge because they may have existing solutions.  JDRF's challenge is with consistency and efficiency with our execution.

patrickmckowen:
@kevinmpowell @ismaeltoure7 I think Ismael's original instinct that the animation with the image would be too distracting is correct.   I suggest removing it; the elevation of the card and the link underline provide the desired level of crafted detail, affordance, and draws the eye nicely to the call to action.

library/components/card_grid/card_grid.scss
```diff
@@ -0,0 +1,243 @@
...
+  margin: 0 auto;
+  padding: 0 $spirit-card-grid-carousel-side-padding;
+  width: 100%;
+
+  @media screen and (max-width: 479px) {
+    display: flex;
+  }
+}
+
+.spirit-card-carousel__button {
```
patrickmckowen: @kevinmpowell should these be Spirit Icon Only buttons?  It appears that they are rectangular instead of round and don't have the spirit focus style: https://gmkr.io/s/5b58f29fa3218a269ae16aa2/0

library/components/card/card.njk
```diff
@@ -0,0 +1,43 @@
...
+        title_text="JDRF One Walk&reg;",
+        body_text="Join more than 900,000 people who come together to raise funds and walk to change the future for the millions of people with T1D.",
+        link_text="Find a Walk Near You",
+        cta,
+        size,
+        aria_label
+        ) %}
+  {% if cta == undefined %}
+    {% set cta %}
+      <span class="spirit-card__link">{{ link_text | safe }}</span>
```
joesnellpdx: @kevinmpowell 

Links is merged into develop. Should the card use `.spirit-link--animated` or duplicate the styles here?

kevinmpowell: I've refactored your link.scss to use mixins and used those same mixins within card. Technically in card the link is the whole card, so the "link" is a `<span>`. I use the mixins to make the "not really a link" look like a link.

PR #179, opened 2018-07-24
Feature/87 link

joesnellpdx:
### Deliver - PR requirmements

* Build into typography sink file
* Update tokens file with necessary style decisions


library/components/link/link.scss
```diff
@@ -0,0 +1,120 @@
...
+  line-height: $line-height;
+  text-align: $text-align;
+  text-indent: $text-indent;
+  text-shadow: $text-shadow;
+  text-transform: $text-transform;
+}
+
+a,
+.spirit-link {
+  @include spirit-typography-reset();
```
joesnellpdx: @kevinmpowell 

Default links should inherit a majority of their font styles -- added the reset so it would work in the sink. 
What is the best practices for handling this here and inheriting font styles except styles explicitly for the link?

joesnellpdx: @kevinmpowell That won't work because `inherit` will inherit all the hostile sink reset styles. Unless, I should crate a dummy `<div>` or `<p>` that gets tome reset styles that get inherited by the link?

Thoughts?

kevinmpowell: Ah, yes I see what you mean. I think link styles that naturally inherit their parent context's typography styles make the most sense.

In this case I think it's fine to add some inline overrides to make the hostile sink behave. I believe there's already a `.hostile-sinki-reset` class or something like that you could wrap them in. Take a look at `forms.njk` for an example.

library/components/link/link.scss
```diff
@@ -0,0 +1,120 @@
...
+  font-weight: $font-weight;
+  letter-spacing: $letter-spacing;
+  line-height: $line-height;
+  text-align: $text-align;
+  text-indent: $text-indent;
+  text-shadow: $text-shadow;
+  text-transform: $text-transform;
+}
+
+a,
```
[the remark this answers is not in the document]
jamesmelzer: Agreed. Plus, we should avoid re-styling the basic attributes within <a> even with the specific class. It's an inline element, even if we make it `display:inline-block`. It should be inheriting the surrounding paragraph's styles for all that, and we should only be overriding the blue color and underline mechanics.

PR #180, opened 2018-07-25
Feature/119 grid

JodiWarren:
- Add prefixed Bootstrap grid
- Add examples from Bootstrap documentation


library/docs/sink-pages/visual-styles/grid.njk
```diff
@@ -0,0 +1,459 @@
...
+    <link rel="stylesheet" href="/styles/grid-sink-page.css">
+    {% filter markdown(true, "hostile-sink-reset") %}
+        # Elevations Sink
+    {% endfilter %}
+
+    {% filter markdown(true, "hostile-sink-reset") %}
+        Basic 3 column grid
+    {% endfilter %}
+    <div class="spirit-container">
+        <div class="spirit-row">
```
joesnellpdx: @kevinmpowell @jamesmelzer 

Do we want macros for rows / columns with grid?

CC: @JodiWarren

library/docs/index.njk
```diff
@@ -1,10 +1,17 @@
...
+  <h2>Components</h2>
   <ul>
     <li>
       <a href="/sink-pages/components/buttons.html">Buttons</a>
     </li>
+  </ul>
+  <h2>Visual Styles</h2>
+  <ul>
+    <li>
+      <a href="/sink-pages/visual-styles/grid.html">Grid</a>
```
kevinmpowell: Move "Grid" under the Components section. That will better reflect the way it's presented in the Doc Site. This is the Grid "component"

library/styles/grid-sink-page.scss
```diff
@@ -0,0 +1,42 @@
+@import '../tokens/tokens';
+@import 'mixins/text-crop';
+
+@mixin space-sink-page-font-reset {
```
kevinmpowell: Just merged this afternoon along with the Forms PR. `spirit-typography-reset` is now available.

library/docs/sink-pages/visual-styles/grid.njk
kevinmpowell: I've seen it done both ways. For "container components" like this, macros don't add a whole lot of value. The "API" for the macro would basically be the ability to pass in a modifier class. I've done similar "container components" already - the `form_field_group` for forms is an example, but I don't think it's necessary at this point.

joesnellpdx:
@JodiWarren 
Approving this branch assuming the following:

- Kevin's update request are made
- Merge conflicts resolve

JodiWarren:
@kevinmpowell I've made those changes and rebased against `develop`.

Are you happy with the directory location of the grid Sass file? It's currently placed it within `styles/mixins`, despite it also being where classes are defined. The problem is that Bootstrap's structure is a bit too interwoven to separate stuff out with a simple copy/paste: https://github.com/JDRF/spirit/blob/a110d41c6185e8ef1bfb2f9fc2b8a594407dcffc/library/styles/vendor/bootstrap/_grid.scss

If you're happy with it then 👍

PR #188, opened 2018-07-27
#46 - Add dark background classes

JodiWarren:
Dark background classes added as per @kevinmpowell's suggestion


library/styles/mixins/background-reverse.scss
```diff
@@ -0,0 +1,18 @@
+@mixin spirit-container-dark-parent-selectors {
```
kevinmpowell: Rename this file to `color.scss`. We may need an equivalent mixin for light parent selectors and other color-related mixins in the future.

library/docs/sink-pages/visual-styles/typography.njk
```diff
@@ -4,7 +4,7 @@
         # Typography Sink
     {% endfilter %}
     <style>
```
jamesmelzer: Would it be worthwhile to move this up out of this file? We could scope it tightly, to `<h3>` right after the hostile sink call.

PR #189, opened 2018-07-27
Feature/20  code example pair

kevinmpowell:
You'll need to run `doc-site` and do a fresh `npm install` there to view this work:
`cd doc-site`
`npm install`
`npm run link-local-library`
`gulp`

Navigate to `/doc-components/code-example-pairs.html` to view the sink page for this doc component. I think I've included examples for most of the common use cases.


jamesmelzer:
Looks great. Love the subtle white text shadow on the code, to amp up the contrast. Approved.

doc-site/docs/doc-components/code-example-pairs.njk
```diff
@@ -0,0 +1,126 @@
...
+
+  {{ esds_doc.code_example_pair(examples=[
+      {
+        example: library.button(text="Learn More"),
+        label: "Default"
+      },
+      {
+        example: library.button(text="Learn More"),
+        label: "On Dark Background",
+        class: "spirit-container--black"
```
joesnellpdx: @kevinmpowell 

This looks / works great. The the default `--dark` should be an option ( jdrf blue ) that works out of the box. Not clear if the intention is for this to be handled in another branch.

kevinmpowell: I believe this will be handled via the work @JodiWarren is doing on containers.

PR #192, opened 2018-07-30
Feature/115 tile

joesnellpdx:
* Kitchen sink file showing all variations in all states, including too much and too little content
* Macro to create all variations
* Update tokens file with necessary style decisions


joesnellpdx:
@kevinmpowell 
@patrickmckowen 
@jamesmelzer 

1. Reverted some over complexity from last week here and assuming we will be using the default grid to display the tiles in the tile-grids page.
2. Found some issues between breakpoints where the tile felt 'too large' but if used a smaller column class felt 'squished.' - see 'Options' below
3. Updated the default grid gutter to 16px ( was set to 32 in the css file ).
4. Medium / large classes exist for the tile - but not applicable when inside a grid 

Options:
1. Use as is
2. Add more granular breakpoints - and update the grid with new column parameters ( add to t-shirt sizes ? ). 
3. Create custom container ( with class options ) for tile grid ( .spirit-tile-grid, .spirit-tile-grid-6up)

library/components/tile/tile.scss
```diff
@@ -0,0 +1,187 @@
...
+    color: inherit;
+    margin: 0;
+    padding: $spirit-space-generic-2-x;
+    text-align: center;
+    text-shadow: $spirit-tile-text-elevation $spirit-tile-text-elevation-color;
+  }
+
+  &:hover,
+  &--hover {
+    box-shadow: $spirit-elevation-5 $spirit-elevation-color-default;
```
jamesmelzer: I recommend adding z-index to the hover state, to pull the active card on top of its peers. In some arrangements, they conflict.

jamesmelzer:
My sense is, editorially, you'll always start with an even row on desktop. In your examples, that's 6, 4, and 3-up. 
On the tile grid sink, I recommend aiming for those golden numbers of tiles to break evenly on all viewports. 

For example, 6@large becomes 4@medium (awkward!) and then 2@small. 3 would be a better medium to maintain the likely case of six cards. 

Suggestion:
6@L | 3@M | 2@S
4@L | 2@M | 2@S
3@L | 3@M | 1@S (!)

jamesmelzer:
Answering your question: I vote for more granular breakpoints, but not linked to the grid. Perhaps...
- Make the largest type breakpoint a little larger - it comes too soon for the 6-up.
- Add an intermediate font size for the 1100-800 range.

library/components/tile/tile.scss
kevinmpowell: I second this, I would expect the hovered card to rise above the others:
screen shot 2018-07-31 at 2 32 00 pm

library/components/tile/tile.scss
```diff
@@ -0,0 +1,187 @@
...
+  position: relative;
+  text-decoration: none;
+  transform: scale(1);
+  transition: box-shadow $spirit-transition-timing-default, transform $spirit-transition-timing-default;
+  width: $spirit-tile-width;
+
+  will-change: transform;
+
+  // When in a grid
+  [class^='spirit-col-'] & {
```
kevinmpowell: I would change this to `.spirit-tile-grid &` so there's a bit more intentionality in switching to this behavior. It's possible a tile may exist within a layout grid somewhere that needs to have one of the three fixed sizes and this prevents that functionality. 

* [ ] Make tile width reactive to the `spirit-tile-grid` class and not the `spirit-col-*` classes

library/components/tile_grid/tile_grid.njk
```diff
@@ -0,0 +1,7 @@
+{% macro tile_grid(
+      class=false,
+      tile_count=1) %}
```
kevinmpowell: Remove `tile_count` if you're not using it. I needed this for generating the card carousel pagination.

* [ ] Remove unused `tile_count` variable

patrickmckowen:
@joesnellpdx I don't have a good answer yet on the sizes and grid; more conversation needed.  

A couple small things:

* The scale size looked good in Sketch but in practice with the motion applied it feels a bit overwhelming.  I think we should dial it down a bit, something like:

```
$spirit-tile-scale-hover: 1.08;
$spirit-tile-scale-hover-md: 1.08;
$spirit-tile-scale-hover-lg: 1.06;
```
-[ ] Update `transform` values

* Ismael's intention with the text positioning inside the card is for it to be vertically centered within the overlay.  To illustrate this better, I've taken a screenshot and marked it up next to the sketch image: https://d.pr/i/BM7UcB

-[ ] Update text alignment within tile

joesnellpdx:
@kevinmpowell @jamesmelzer @patrickmckowen 

All requests have been updated except:

Grid: cannot simply alter the grid for tiles. Either will need to do a custom grid, or modify the existing grid ( specifically about adding breakpoints - that will bring up the t-shirt size question of what do we call a breakpoint between md and lg? ). Another option is to reduce the text sizes, but that doesn't totally solve the problem.

Next:
Shall I merge and we can alter / update the grid portion later, or keep open until we resolve?

patrickmckowen:
My preference is to alter Tile to work within the updated grid+container approach I shared this week.  I'd like to play with it in code, rather than Sketch, to make recommendations for how Tile may need to be adjusted for mediumish screen sizes.  I'm not sure what that means in terms of branching strategy.

joesnellpdx:
Based on @patrickmckowen 's comment, I'd suggest we merge into develop. Test with the updated container approach, then see if it needs tweaking.

jamesmelzer:
Sounds good, @joesnellpdx

Issue #190, opened 2018-07-30
Tell the linters to ignore Bootstrap code


jamesmelzer:
The Bootstrap grid code is throwing tons of linting errors in /library/. Get the linters to ignore it.

PR #194, opened 2018-07-31
#190 - Don't lint vendor files

JodiWarren:
Quick fix to remove the sass-lint errors on vendor style files, specifically the Bootstrap files.


JodiWarren:
@jamesmelzer @kevinmpowell as requested, linting disabled on these files.

PR #195, opened 2018-07-31
Feature/91 long form text

JodiWarren:
@kevinmpowell @jamesmelzer Please take a careful look at the long form content.

At the moment I've mostly stuck quite closely to the Sketch file. I've added a couple of right-hand oriented layouts for the pull quote and for the floated images. If you don't want those, then I can simplify things a bit.

### DIFF
--- a/library/components/body-text/body-text.scss
+++ b/library/components/body-text/body-text.scss
@@ -6,95 +6,104 @@ $spirit-spacing-extra: $spirit-spacing-base * 1.5;
 $spirit-spacing-double: $spirit-spacing-base * 2;
 
 .spirit-body-text {
+  @include spirit-typography-reset;
+
+  h1 {
+    @include spirit-h1;
+    margin-top: $spirit-spacing-title;
+    margin-bottom: $spirit-spacing-base;
+  }
+
+  h2 {
+    @include spirit-h2;
+    margin-top: $spirit-spacing-title;
+    margin-bottom: $spirit-spacing-base;
+  }
+
+  h3 {
+    @include spirit-h3;
+    margin-top: $spirit-spacing-extra;
+    margin-bottom: $spirit-spacing-base;
+  }
+
+  h4 {
+    @include spirit-h4;
+    margin-top: $spirit-spacing-base;
+    margin-bottom: $spirit-spacing-base;
+  }
+
+  h5 {
+    @include spirit-h5;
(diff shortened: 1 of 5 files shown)


patrickmckowen:
@jamesmelzer I don't think the max-width is something that should be baked into the long form component.  It's guidance that designers will use when constructing layouts.

jamesmelzer:
Ah, okay. That simplifies things.

JodiWarren:
@jamesmelzer You removed a few items from the sink page, like the quotes and the images/videos. Does that mean that they're not required? I'll happily trim down the CSS if that's the case.

JodiWarren:
@jamesmelzer Good point on the namespace. I've renamed the class to `spirit-long-form-text` just for consistency. I'm happy to switch it all to `spirit-body-text` if you prefer the shorter name.

jamesmelzer:
Weird. That was unintentional. Restoring and reviewing now...

library/components/long-form-text/long-form-text.scss
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  img {
+    max-width: 100%;
+    display: block;
+    margin-top: $spirit-space-generic-2-x;
+    margin-bottom: $spirit-space-generic-4-x;
+  }
+
+  .spirit-image-left,
```
jamesmelzer: Would these be better BEM notation as .spirit-image--left, .spirit-image--right?

library/components/long-form-text/long-form-text.scss
```diff
@@ -0,0 +1,212 @@
...
+  .spirit-image-right {
+    @include float-item;
```
jamesmelzer: Consider adding nested `img {margin-top: 0;}` to negate the double top margin when wrapping an image and caption with this.

jamesmelzer:
The *images* and *blockquotes* look great. I like your spacing instincts on the images. I think this is the correct place to define these. In the future, if we do something more involved (linked images that open into a lightbox, for example) then I'd move that into a new component.

library/components/long-form-text/long-form-text.scss
JodiWarren: @jamesmelzer I didn't do these as BEM modifiers because I can't guarantee that they'll also have the `.spirit-image` class. I thought it best to assume behaviour on all images, and then allow the author to add an optional float.

PR #199, opened 2018-08-02
Feature/32  token table doc component

kevinmpowell:
Another `npm install` for this one :( sorry!

`git checkout feature/32--token-table-doc-component`
`cd doc-site`
`npm install`
`npm run link-local-library`
`gulp`

Two sink pages for review. Nunjucks comments have usage details:
`/doc-components/reference-tables.html`
`/doc-components/token-tables.html`


jamesmelzer:
The right/left border you added to the headers is an artifact in Sketch (stupid Sketch!). No borders intended. I added a commit to remove them. 

Separately, am I correct that the token table doesn't have a setting for column widths? What do you think about passing a default (e.g. 75, 50) or letting us set that? In your examples, that right column is all over the place. 

Otherwise, looks great. Approved.

kevinmpowell:
> Separately, am I correct that the token table doesn't have a setting for column widths?

I'll add one.

PR #202, opened 2018-08-03
Feature/196  tint stack reverse

kevinmpowell:
@jamesmelzer I ended up deleting the custom color page macro I'd created. It reduced some boilerplate but was limiting the ability to customize that the base `esds_doc.tint_stack` macro has. You _will_ need to do an `npm install` for this one.

* checkout branch
* `npm install`
* `npm run link-local-library`
* `cd doc-site`
* `gulp`

Page can be viewed at: /visual-style/color.html


jamesmelzer:
Hey, and you even fixed my color page along the way. Looks great.

PR #208, opened 2018-08-09
#15 Add Do/Don't Doc Component

kevinmpowell:


jamesmelzer:
@kevinmpowell I thought we decided not to use the icon-only style of Do/Don't; instead we're sticking with boxes all the time. (I also thought I had purged them all, but I see **a pair slipped through** in the "doc components" artboard. Sorry about that.)

kevinmpowell:
@jamesmelzer 🤦‍♂️ I had suggested it, but saw them in the dartboards and made an assumption instead of asking. Trivial to remove.

Issue #167, opened 2018-07-09
Need to audit utilities.scss and remove styles that are covered by other component .scss files


kevinmpowell:

kevinmpowell:
`utilities.scss` has been removed.

PR #203, opened 2018-08-06
Feature/176 container

kdo:
Hello!
This is my first commit, hopefully, I'm not too far away from the design system "spirit" 😅 

- Creates a new `Container` component
- Adds `Container` sink page with examples (fullwidth vs. large, padding top & bottom options, background color options, stacked containers and containers containing the grid) I've added an explicit pink background color on the demo containers to show the padding.

Thanks!

PR #210, opened 2018-08-10
#167 move icon styles into icons.scss, delete container styles since …

kevinmpowell:
This is pretty straightforward. @joesnellpdx you may want to pull this branch into your containers branch and make sure nothing in the containers sink was depending on the container classes I deleted from `utilities.scss`


jamesmelzer:
Agreed. I'd still expect to have basic helper classes for `.spirit-container--brand` and similar. If we don't really need them in the library, let's move them into the doc site.

kevinmpowell:
@jamesmelzer I agree, I imagine those will be recreated in `/library/components/container/container.scss`

joesnellpdx:
@jamesmelzer @kevinmpowell 

I have questions on that in my PR:
https://github.com/JDRF/spirit/pull/203#issuecomment-411887245

PR #225, opened 2018-08-22
Feature/44 space pub

jamesmelzer:
I added two helper mixins to `library/styles/helpers/spacing.scss` to add `inset` and `inset-squish` to our list of space helper classes.


jamesmelzer:
@patrickmckowen Updated with square corners.

jamesmelzer:
@patrickmckowen When you get a chance, I could use your review on this (space) doc.

patrickmckowen:
@jamesmelzer I have one question about token naming which we can separate from this PR and one question about the doc itself.

Is it intentional that the on-page nav [splits the two opening paragraphs](https://gmkr.io/s/5b84377f356b1d1f98f909ab/0)?

jamesmelzer:
Good Q. 
The title and deck paragraph are a consistent block at the top of all articles. What comes "next" in content is up to the guideline author - it could be a heading, illustration, or paragraph. 

Current state
* Right now, on moderate-sized viewports, anchor nav injects itself in between title/deck and the rest of the content. 

Expected future state
* In the future, once page shell+hamburger menu are ready, I'd like to collapse the left nav instead of injecting right nav like this. That will redouble the reader's focus on the content, and re-enable the anchor nav's stickiness in the sidebar. 

Thoughts?

patrickmckowen:
@jamesmelzer okay, that makes sense and I think this is ready to merge.

Sounds like some more convo is needed around further refining the the structure and layout of the page.  TBD.

PR #227, opened 2018-08-22
Feature/177 layout grid build

joesnellpdx:
- Current ticket responsibilities: Adding brush top  / bottom and brush background options
- Display new options across breakpoints with different content


joesnellpdx:
@kevinmpowell 

I need a sanity check on this one before I continue with implementation:

- Macros
- The way I'm handling the brushes ( paint and bg )

Once we can agree on implementation, will add some more content options.

kevinmpowell:
> Macros

I'm a little confused by `background_color`, `inner_background_color`, `paint`, and `brush_bg`. It's difficult for me to tell what combinations do what. Some of them seem color/brush related, others seem related to position, so I think some clearer names for those arguments are warranted. You also have a typo `conatiner`.



> The way I'm handling the brushes ( paint and bg )

This seems fine. I've never dealt with inline SVGs in SCSS like this. Can they be referenced as external files? Or if not, moved into scss variables into a separate file, just so they don't dominate this file. A variable name or external file name would make it clearer which brush was being included.

Was there anything else specific you wanted me to take a look at?

PR #232, opened 2018-08-24
Feature/223  cdn asset changes

kevinmpowell:
You'll need to `npm install` in `library` before running `gulp` in order to build the minified version of the stylesheet.
I also updated some of the build config for `doc-site` so the fonts will reside at the correct path for CDN consumers.


jamesmelzer:
I too built locally and poked around in the resulting `library/dist/`, `/docs/`, and `/doc-site/dist/` folders. Looks good.

PR #233, opened 2018-08-24
Add CSS properties to support icons in IE11

kevinmpowell:
## Working in IE11 Now
Note the dots on exclamation points and question marks.
screen shot 2018-08-24 at 4 16 44 pm

Added some CSS properties that are inlined in our sprite. Since IE11 relies on a polyfill that injects the SVG content inline, those inline properties were getting stripped out. Fortunately they're targetable via CSS.

### DIFF
--- a/library/components/icon/icon.scss
+++ b/library/components/icon/icon.scss
@@ -5,6 +5,9 @@
   height: 1em;
   line-height: 1em;
   stroke: currentColor;
+  stroke-linecap: round; // Set inline on the SVGs but stripped by the polyfill we use for IE11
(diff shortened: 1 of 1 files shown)


library/components/icon/icon.scss
```diff
@@ -5,6 +5,9 @@
   height: 1em;
   line-height: 1em;
   stroke: currentColor;
+  stroke-linecap: round; // Set inline on the SVGs but stripped by the polyfill we use for IE11
```
joesnellpdx: @kevinmpowell Would not have anticipated this to be the solution... great fix and glad it works!

PR #236, opened 2018-08-28
Fix/missing icons

kevinmpowell:
Should we include all the icons in `library`? We've gone from `8kb` to `80kb` - discuss!


jamesmelzer:
Okay, I'll get us started. 

The current selection (27) is very tight so it's reasonable to expect that to double to ~60 icons. In my experience, 60 is a pretty generous set of UI icons. If we're not doing things like rich text editor toolbars or dozens of file type icons, 60 should cover a broad range of scenarios. 

(FYI the full Feather collection is 266 icons)

jamesmelzer:
@patrickmckowen Good list.

kevinmpowell:
I've moved forward assuming we want to surface all of them, if we want to reduce the number shown on the doc site that's an easy fix.

patrickmckowen:
@kevinmpowell 
Looks good to me 👍

PR #238, opened 2018-08-28
Release/#237 1.0.0 alpha.1

joesnellpdx:
- Library release to NPM


kevinmpowell:
@joesnellpdx I'm realizing we need some way to differentiate if this is the `library` part of the release process or the `doc-site` part of the release process. Can you call that out in the PR description?

I'm assuming this is the `library` release to NPM. If that's the case, I approve, I do have some questions when we get to the `doc-site` release.

PR #243, opened 2018-08-30
Feature/62 form pub

jamesmelzer:
Actions/decisions needed:
- [x] @kevinmpowell to look at attributes field on radio and checkbox in .njk - is it passing a trailing slash "/" ? 
- [x] Decide what to do with solo checkbox with a validation error (last example in the validation section). Two options shown.


kevinmpowell:
@jamesmelzer no trailing slash, that's part of the self closing `/>` from the end of the `<input />` tag. I've made that consistent across all the self-closing form inputs now.

jamesmelzer:
I constrained certain input examples to 80% of the width of the example block. It's a page-specific style, and has to be called specifically on each code/example block we want it on (rather than changing them all page-wide or site-wide).

jamesmelzer:
Added input masks in the input section.

jamesmelzer:
@joesnellpdx Am I right in assuming that if a product has `spirit.js` loaded, masked inputs will work as advertised without further hooks in code? If so, I don't think further documentation is warranted. 

We could add a note at the top of the input masks section reiterating the JS prerequisite.

joesnellpdx:
@jamesmelzer Spirit.js does not include dependencies. See `/getting-started.html` in develop - has new(ish) updates under the Javascript Section.

Spirit.js never intended to have npm packages. Part of why we enhanced this documentation and added the bundle and dependencies options.

CC: @kevinmpowell

jamesmelzer:
> Spirit.js never intended to have npm packages. Part of why we enhanced this documentation and added the bundle and dependencies options.

Ah, I misunderstood. So mentioning JS at all opens up a major can of worms. On Form, can we say something like "Masking depends on Javascript. See Getting Started for Developers for information on the necessary dependencies."

joesnellpdx:
@jamesmelzer That sound perfect!

PR #244, opened 2018-09-04
Feature/19  typography swatch

kevinmpowell:
Sink page for the doc component is at:
`/doc-components/typography-swatches.html`

@jamesmelzer it should have all the options you need for the long form text and typography doc pages.


jamesmelzer:
It took me a while to parse all the options, but this looks great!

PR #245, opened 2018-09-05
Feature/127  page shell

joesnellpdx:
**Page shell:**

- Includes off-canvas element - demo nav menu - placeholder
- Includes body, main, footer, and site wrappers
- Includes some containers for demonstration purposes

View in library:
`/components/shells.htm`


library/components/shell/shell.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Spirit Shell Scripts
+ *
+ * @return {void}
+ */
+const spiritShell = () => {
```
joesnellpdx: @kevinmpowell 

Can I add a class to the html element with the block in the .njk file? Would prefer not having to add in the library via js, but this works.

library/components/shell/shell.njk
```diff
@@ -0,0 +1,51 @@
+{% macro shell(
+        body_class=false,
+        body_modifier=false,
+        main_class=false,
+        main_modifier=false
+        ) %}
+        <html class="test">
```
kevinmpowell: What's the rationale for including `<html>` and `<body>`?
In my past shell work I've always built this as a component that lives within `<body>` and doesn't rely on anything outside of a containing `<div class="spirit-page-shell">` wrapper.

joesnellpdx:
@kevinmpowell 

Moved to self contained elements in this commit:
https://github.com/JDRF/spirit/pull/245/commits/4bc1efdcaf25444bdc9a504fd1bafc963d470278

library/components/shell/shell.njk
```diff
@@ -0,0 +1,45 @@
+{% macro shell(
+        class=false,
+        modifier=false
+        ) %}
+      <div id="spirit-shell" class="spirit-shell{{' spirit-shell--' + modifier if modifier }}{{ ' ' + class if class}}">
```
kevinmpowell: Are the `id` attributes needed? They're mirrored by the default `class` so I'm wondering what they're being used for.

* [ ] Remove `id` attributes if not needed.

library/components/shell/shell.njk
```diff
@@ -0,0 +1,45 @@
...
+  %}
+    <header id="spirit-shell__header" class="spirit-shell__header{{ ' ' + class if class}}{{' spirit-shell__header--' + modifier if modifier }}">
+      {{ caller() }}
+    </header>
+{% endmacro %}
+
+{% macro shell_main(
+      class=false,
+      modifier=false,
+      article_id="spirit-shell__article",
```
kevinmpowell: I don't see where `article_id` `article_class` or `article_modifier` are being used.

* [ ] Remove macro arguments if not needed

joesnellpdx:
I need at least two of you to review following commit per today's slack discussion:
https://github.com/JDRF/spirit/pull/245/commits/760e980c6985786331d89b561e9c52b88bdde499

- [ ] @patrickmckowen 
- [ ] @jamesmelzer 
- [ ] @kevinmpowell

joesnellpdx:
@kevinmpowell Updates made

library/styles/mixins/accessibility.scss
```diff
@@ -12,33 +12,9 @@ $spirit-screen-reader-text-line-height-xs: 1.1429;
...
       clip: auto;
-      color: $spirit-interactive-color-text-primary;
-      cursor: pointer;
-      display: inline-block;
-      font-family: $spirit-font-family-sans-serif;
-      font-size: $spirit-font-size-s;
-      font-style: normal;
-      font-weight: $spirit-font-weight-bold;
-      height: auto;
       left: $spirit-space-inset-1-x;
```
kevinmpowell: Setting `left` and `top` limits reusability of this mixin. I would expect:
```
clip: auto;
left: auto;
position: static;
z-index: auto;
```

To effectively just "reset" what the mixin does typically on focus. If other positioning is desired on focus that should be handled in context, not in the mixin.

kevinmpowell:
@joesnellpdx looks good 👍

joesnellpdx:
@jamesmelzer this is done

PR #253, opened 2018-09-06
WIP LFT revision and compare sink

jamesmelzer:
I created a side-by-side comparison page called long-form-text-2.html that shows original and revised LFT. 

What changed? 

- All LFT top margins set to 0 (removes the default spacing in the reset file)
- Set the intentional top spacing using padding, not margin (on headings, mostly)
- Any descending heading combination (e.g. h2 + h3) sets the top padding to zero. This works like margin collapse, but more predictably in a fixed set of scenarios.
- List items are now the same color as body text.

### DIFF
--- a/library/components/long-form-text/long-form-text.scss
+++ b/library/components/long-form-text/long-form-text.scss
@@ -226,3 +226,264 @@ $spirit-space-generic-2-x-half: 18px;
     margin-top: $spirit-space-generic-2-x * -1;
   }
 }
+
+
+.spirit-long-form-text-2 {
+  @include spirit-typography-reset;
+  @include clearfix;
+
+  h1 {
+    @include spirit-h1;
+    margin-bottom: $spirit-space-generic-2-x;
+    margin-top: 0;
+    padding-top: $spirit-space-generic-5-x;
+  }
+
+  h2 {
+    @include spirit-h2;
+    margin-bottom: $spirit-space-generic-2-x;
+    margin-top: 0;
+    padding-top: $spirit-space-generic-5-x;
+  }
+
+  h3 {
+    @include spirit-h3;
+    margin-bottom: $spirit-space-generic-2-x;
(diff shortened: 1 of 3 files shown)


library/components/long-form-text/long-form-text.scss
```diff
@@ -1,7 +1,7 @@
 // Possible token
 $spirit-space-generic-2-x-half: 18px;
```
patrickmckowen: @jamesmelzer wouldn't 2-x-half be 8+8+4=20?

patrickmckowen: @jamesmelzer wouldn't 2-x-half be 8+8+4=20?

library/components/long-form-text/long-form-text.scss
```diff
@@ -1,7 +1,7 @@
...
```
kevinmpowell: * [ ] Remove `spirit-long-form-text-classic` class

jamesmelzer:
Final changes after visual approval:

- [ ] Removed `.spirit-long-form-text-classic`
- [ ] Made the side-by-side comparison the default LFT sink and deleted the old one. I updated the sink index.njk to reflect this change. 
- [ ] Changed the side-by-side comparison to show default and brand backgrounds (instead of original versus revised LFT)
- [ ] Changed the side-by-side comparison's "classic" side to use the new LFT, so both sides are identical LFT, just on light or dark background
- [ ] Removed the non-standard 18px variable `$spirit-space-generic-2-x-half` and reset it back to 16px. If you approve, @patrickmckowen, this change will need to be made in the Sketch sticker sheet, too. Note: This was used as the bottom-margin below lists in `List` and `Long Form Text`. I updated both in code. 

I think a second code check is appropriate, @kevinmpowell. Friday is fine.

kevinmpowell:
@jamesmelzer looks good.

Issue #131, opened 2017-08-29
Breakpoints


jamesmelzer:
### Deliver

* Prototype full combination of page shell, layout containers, and grid with a sample of real components. (@joesnellpdx has already set much of this up for the containers ticket).
* Quick audit of breakpoint usage across components, just to make sure there's no surprises. (container vs tokens file)
* Test across a limited sample of real devices. Recommend adjustments to breakpoints (shift up or down). SHARE: for other folks on the team to try, too. 
* Define or redefine a mixin for setting media query to one of the pre-defined breakpoints
* Update tokens file with necessary style decisions

PR #262, opened 2018-09-11
feature/131--breakpoints: #131 add breakpoints sink page with samples

joesnellpdx:
- Setup sink page with samples
- Use to test breakpoints / xbrowser
- Use to share with other stakeholders if needed


joesnellpdx:
@jamesmelzer @patrickmckowen @kevinmpowell 

Initial sink page setup:
`sink-pages/components/breakpoints.html`

Are there any specific combinations you would like to be added or modified?

jamesmelzer:
Oh, @joesnellpdx I love this sink page! 

I'll give it a more thorough review after critique. Thanks for putting this together!

kevinmpowell:
> Oh, @joesnellpdx I love this sink page!

@jamesmelzer built this one long ago. He loves those breakpoint testing pages: https://eightshapes.com/responsive-grid-demo/

kevinmpowell:
@joesnellpdx looks great. Only bit of feedback I have is consider renaming it and possibly moving it from `library` to `doc-site`. You're testing a lot more than just breakpoints. This is almost a "master sink" page that demonstrates component composition. I wonder if this is close to a "sample template page." 

I could see a collection of these demos (article, search results, dashboard, homepage) being valuable as part of the doc site.

joesnellpdx:
@kevinmpowell I can see moving to the doc site - but I'd like to have the navigation component and hero in place first - so we aren't using 'dummy' components. Does that make sense?

Renamed to 'page'

@patrickmckowen Please review to see if there are any combos or updates needed.

patrickmckowen:
@joesnellpdx this is super helpful.

One request:
- [ ] Please add Cards to Carousel on mobile and Cards to List on mobile

One (extremely minor) issue that I bring it up in case it's relevant to the convo we had with Ross today about Container gutters:
- [ ] Between 1048px and 1080px, the brush background doesn't reach the edge of the screen: https://gmkr.io/s/5b99acd6595b7048707f0275/0

joesnellpdx:
@patrickmckowen 

Added the two components - see the following 'titles':
- Cards List to 4up
- Cards Carousel to 4up

Also, fixed space on the sides of the wave backgrounds - had not accounted for grid gutter - good catch!

@kevinmpowell I think adding the 'page' sample to the doc site is a good idea after thinking about it. I attempted adding and not sure the best way to add a 'sample page' -- I think we'd want it without any of the doc site shell. Good thing I attempted, because I'd like to resolve this before we disconnect.
- how do I create a template without the doc site stuff ( or should i)?
- `{% call library.shell() %}` throws errors - even after  linking local library - how do I get this to work?
- What is the suggested pattern for custom styles / images / assets for a sample page (i.e. demo components that are in the library sample page)?
- any other tips or suggestions for this?

jamesmelzer:
I don't have any substantial feedback. I like Kevin's idea of moving this into the doc-site space, so it gets published as a model.

jamesmelzer:
@kevinmpowell 

- [ ] Answer Joe's questions above
- [ ] Create a "hello world" page on the doc-site side, so Joe has a starting point

But...

Decision: we're not going to *publish* example pages until Patrick/Ismael have designed them.

kevinmpowell:
> Answer Joe's questions above
> [ ]  Create a "hello world" page on the doc-site side, so Joe has a starting point

@joesnellpdx have set up a meeting to walk through this process next week. I recommend moving forward with this PR and merging and we can add the additional work after that meeting as a separate task.

jamesmelzer:
@kevinmpowell Joe's technically off the project next week, so we were hoping you could take the next step partial step and then leave it open for 10up to continue later. I think the next step is: set up a minimal template in doc-site with appropriate includes (e.g. CSS), but not any doc site layout or components (e.g. no main nav). Include one hello world component, like a button, to confirm styles are working and to provide a syntax example.

PR #263, opened 2018-09-12
feature/56 Icon Pub

patrickmckowen:
https://docs.google.com/document/d/13-xUlX8sBk_qdTyFOzZraG4r-Eh_7fj6PZk4q28tmMk/edit# 

Two issues, both with Color Tokens table > On Dark row:

- [x] The doc references a token I couldn't find in the library: `$spirit-icon-color-on-dark`  I used text-color instead.
- [x] I couldn't figure out how to change that row's background color.

Related, I didn't see a Tint Stack file in doc-components and so I pulled an example from the Color doc page.


jamesmelzer:
Huh. I guess we never created that token. Perhaps we should, now? It's used for things like the specialized buttons, which currently refer to the palette token directly (`$spirit-color-white`) e.g.

<code>
  @include spirit-container-dark-parent-selectors {
    background-color: $spirit-feedback-color-foreground-danger;
    box-shadow: $spirit-elevation-1 transparent, inset 0 0 0 $spirit-border-width-button-secondary $spirit-feedback-color-foreground-danger;
    color: $spirit-color-white;
  }
</code>

jamesmelzer:
Reviewed. Looks good. Boy does this make me want to roll out the LFT update to get bullets that match the paragraphs...

For the future, add `color_role: 'foreground'` to a tint stack item to reverse it onto black. I'm committing that change now.

doc-site/docs/visual-style/icons.njk
```diff
@@ -1,14 +1,131 @@
...
+
   {% call esds_doc.icon_swatch_grid(shims=6) %}
     {% for icon in spirit_icons %}
       {{ esds_doc.icon_swatch(class="spirit-doc-icon-swatch", name=icon) }}
     {% endfor %}
   {% endcall %}
+
+  {% filter markdown %}
+    ### Using Icons in Code
+    Reference icons from a layered sprite file, `spirit.svg`, which loads and caches all Spirit icons in a single resource call. Use the following syntax:
```
kevinmpowell: Trivial content change (and sorry if I missed the doc review process for this one) but I've never heard an SVG sprite referred to as "layered." I'm not sure where that term is coming from. Consider removing.

* [ ] Consider removing "layered" from description of sprit file

kevinmpowell:
> For the future, add `color_role: 'foreground'` to a tint stack item to reverse it onto black. I'm committing that change now.

Shouldn't all of these be reversed?
screen shot 2018-09-12 at 1 53 04 pm

doc-site/docs/visual-style/icons.njk
```diff
@@ -1,14 +1,132 @@
...
+  {% filter markdown %}
+    ### Class Reference
+  {% endfilter %}
+
+  {{ esds_doc.reference_table(column_widths=[25, 25, 50],
+    data=[
+      {
+        class: '.spirit-icon',
+        applies_to: '<svg>',
+        outcome: 'Applies secondary button style'
```
kevinmpowell: I'm not sure if this is a copy paste fail or not but I don't expect to see "Applies secondary button style" in the description of an icon-related class.

I'd remove this class reference altogether. I recommend documenting only modifier classes. The required classes are intrinsic in the copy/paste code examples elsewhere on the page.

* [ ] Consider removing class reference table

jamesmelzer: That's all me. How else would you describe it? the icons are all on top of one another, rather than a traditional sprite where they're spread out in a two-dimensional matrix. 

Maybe it doesn't matter, and we just remove it? I'd be okay with that.

kevinmpowell: I vote for removal. I haven't seen that term in any of the articles I've referenced

Issue #280, opened 2018-09-25
Spirit Button Icon - incorrect size and size modifiers not working


joesnellpdx:
- Make sure the spirit button icon matches Abstract / Sketch (both the icon and the button size): https://share.goabstract.com/d4d9f3e8-1ef0-4571-8c8d-e94e64858f2c
- Test sizing modifiers - make sure match sketch / abstract
- Test context ( background colors )

joesnellpdx:
In Peer review:
https://github.com/JDRF/spirit/pull/300

ismaeltoure7:
@joesnellpdx I've reviewed this but I am not listed as approver on the merge request.

joesnellpdx:
This has been merged into develop.

Will also need to verify doc site is updated correctly after this is merged (new github ticket)

NOTE:
for all of these updates - will need method to test / remove 'fixes' on .org when next release is pushed.

In this instance:
Need to review cases of .spirit-button--icon-only in css and in codebase output for updates / alterations on .org. In particular - in .social-nav

PR #300, opened 2019-01-04
feature/280-spirit-button-icon-size: #280 - fix icon button sizing

joesnellpdx:
Fix issue with attribute selector being `starts with` rather than `contains` (not targeting svg class appropriately)

Review in sink:
/sink-pages/components/buttons.html

Will also need to verify doc site is updated correctly after this is merged (new github ticket)

NOTE:
for all of these updates - will need method to test / remove 'fixes' on .org when next release is pushed.

In this instance: 
Need to review cases of `.spirit-button--icon-only` in css and in codebase output for updates / alterations on .org. In particular - in `.social-nav`

### DIFF
--- a/library/components/button/button.scss
+++ b/library/components/button/button.scss
@@ -151,7 +151,7 @@ $spirit-space-inset-squish-6-x-negative: -24px -48px !default;
   top: 0;
   width: 100%;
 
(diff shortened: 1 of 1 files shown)


ismaeltoure7:
Looks good to me!

Issue #299, opened 2019-01-03
Checkbox issue on iOS


patrickmckowen:
From Jamie Arkin at Blue State Digital:

As you can see in the [attached] video, when you click the checkbox it doesn’t check. It seems like it goes into a focus state of some sort. You are required to click twice in order for the checkbox to work, then it seems to work perfectly fine.

[spirit-checkbox-error.mov.zip](https://github.com/JDRF/spirit/files/2724843/spirit-checkbox-error.mov.zip)

Jamie suggests this fix:

```
var dragging = false;
$('body').on('touchmove', function() {
	dragging = true;
});

$('body').on('touchstart', function() {
	dragging = false;
});

$('body').on('click, touchend', 'label', function(e) {
	e.preventDefault();
	if (dragging) return;

	var input = $(e.currentTarget).find('input');

	if (input.is(':checked')) {
		input.prop('checked', false);
		return;
	};
	input.prop('checked', true);
});
```

joesnellpdx:
@patrickmckowen @lauralvictor 

PR here:
https://github.com/JDRF/spirit/pull/303

joesnellpdx:
This has been approved and merged

PR #303, opened 2019-01-09
feature/#299-checkbox-nocheck-xbrowser:

joesnellpdx:
#299: move pseudo border for checkbox and rado to default of element and change color on hover

Weird one! Just the location of the pseudo element for the 'hover' border state was disrupting the 'checked' state. Moved to default transparent border and just changing color on hover - and it works!

Please verify on iOS and ie11.

`/sink-pages/components/forms.html`


joesnellpdx:
@ismaeltoure7 added you as reviewer here. Thought I had done last week.

PR #333, opened 2019-02-12
feature/#320-button-block-level

joesnellpdx:
Issue request is to make block level button option. After investigation, in-line buttons and block level buttons collapse margins differently. Therefore, created a new class option for `spirit-button--fullwidth` to give 100% width to the standard inline spirit button. 

I also added the class reference to the button page in the doc site. I did not, however, update any further documentation or use of this class in the button doc page. If this is required, will need further direction or recommend adding new issue to do so. 

* have to gulp in library
* for doc site ( after library compiled ):
  -- `npm run unlink-local-library`
  -- `npm run link-local-library`
  -- `gulp`

See: 
`/sink-pages/components/buttons.html` in library
`/components/buttons.html` in doc site


joesnellpdx:
@ismaeltoure7 @lauralvictor this is ready for review.

PR #344, opened 2019-02-14
feature/#343-card-list-card-truncate: add css truncation

joesnellpdx:
Ismael recommended I add this as a separate PR for internal discussion. 

CSS added to truncate text to two lines.

Issues:
Not supported cross browser. Fallback - only two lines, but no ellipsis.

Q: Thoughts on this implementation? Suggestions?

I feel it is a bit hacky and prefer not doing this, but it should work in _most_ mobile browsers and we are only applying below our mobile breakpoint.

### DIFF
--- a/library/components/card/card.scss
+++ b/library/components/card/card.scss
@@ -35,7 +35,7 @@
   }
 
   .spirit-card__body {
-    display: none;
+    margin-top: -$spirit-space-inset-half-x;
   }
 
   .spirit-card__text {
@@ -78,6 +78,19 @@
       $font-size: $spirit-font-size-s
     );
     margin: $spirit-space-stack-2-x;
+
+    @media screen and (max-width: $spirit-breakpoint-m) {
+      // Truncate body text on small screens
+      // sass-lint:disable no-vendor-prefixes property-sort-order
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: pre-wrap;
+      display: -webkit-box;
+      -webkit-box-orient: vertical;
+      -webkit-line-clamp: 2; // Number of lines you wish to display
+      max-height: 42px; // Number of lines times line height
+      // sass-lint: enable
+    }
   }
 
   .spirit-card__link {


joesnellpdx:
@ismaeltoure7 , @jamesmelzer and @kevinmpowell : I'd love your take here while @patrickmckowen is out.

library/components/card/card.scss
```diff
@@ -78,6 +78,19 @@
...
+    @media screen and (max-width: $spirit-breakpoint-m) {
+      // Truncate body text on small screens
+      // sass-lint:disable no-vendor-prefixes property-sort-order
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: pre-wrap;
+      display: -webkit-box;
+      -webkit-box-orient: vertical;
+      -webkit-line-clamp: 2; // Number of lines you wish to display
+      max-height: 42px; // Number of lines times line height
```
kevinmpowell: At a minimum I'd calculate this based on line height tokens, so it's not a magic number.

kevinmpowell:
@joesnellpdx can you add a sink page example that demo's this behavior?

joesnellpdx:
@kevinmpowell 
- used token for line height
- created sink page: '/sink-pages/visual-style/truncation.html' - view below 768px to see truncation

library/components/card/card.scss
```diff
@@ -110,7 +110,25 @@ $spirit-card-list-text-box-shadow: inset 0 -1px 0 0 $spirit-interactive-color-se
...
+      max-height: #{$spirit-font-line-height-body * 2}em; // Line height * number of lines
+      // sass-lint: enable
+    }
+=======
     margin: $spirit-space-stack-1-x;
```
kevinmpowell: @joesnellpdx git conflict markers are preventing this from compiling.

kevinmpowell:
@joesnellpdx I removed the git conflict markers locally so this would build. It looks good in the browser, as long as the missing ellipses in Firefox are acceptable from a design perspective I don't see any other issues with this.

Doesn't feel that "hacky" to me.

joesnellpdx:
@patrickmckowen @ismaeltoure7 

Will wait for your assessment and approval of this before moving forward.

joesnellpdx:
@ismaeltoure7 

Currently, truncation is on all cards on smallest screens. Should I change cards in list view only on small screens - or leave as is for now to review?

ismaeltoure7:
@joesnellpdx please leave as-is for review, as the only cards which would not be affected are the Stack and Carousel cards.

joesnellpdx:
Per team discussion, we are not implementing this update at this time. However, we are keeping this MR open for reference.

PR #355, opened 2019-02-21
Feature/#336 ie11 radio field label wrapping

joesnellpdx:
Give checkbox group label a max width of 100% to not allow overflow on ie11


lauralvictor:
@joesnellpdx I see the "Group Horizontal Too Much Content" section shows the text overrunning  the screen instead of wrapping, and a scrollbar displays to allow viewing the entire text.
bs_win10_ie_11 0 8

joesnellpdx:
@lauralvictor 

Great catch! I pushed a fix. Please review.

lauralvictor:
@joesnellpdx Thanks, that fix did it.

joesnellpdx:
@lauralvictor Can you clarify? I'm not seeing the issue in your screenshot or when inspecting.

The Group horizontal - should be 'inline' so the `yellow` and `orange` should be next to each other. The 2nd in that group doesn't apply because it is full width (too much content). I'm not sure if that is what you are looking at or not?

lauralvictor:
@joesnellpdx Thanks for that clarification. What I'm seeing does match your description.

PR #358, opened 2019-02-22
Feature/#316 list object fit refactor

joesnellpdx:
In updating the images for card and list, I found some inconsistencies and an opportunity to componentize the image in a macro to have better control now and in the future for these images.

- **includes card update branch** - https://github.com/JDRF/spirit/tree/feature/%23315-card-object-fit-refactor - as not yet approved + merged at the time of this MR
- add object fit macro
- add media sink - for testing images
- add object fit macro to cards and lists
- update card and list styles for the object fit macro
- update card and list sink pages / macros to use object fit macro

To test:
View the following pages on modern browsers and IE 11. The list and card pages, view on mobile and desktop
- `/sink-pages/components/card-grids.html`
- `/sink-pages/components/list-grids.html`
- `/sink-pages/components/images.html`


joesnellpdx:
@kevinmpowell 

If you still have bandwidth, I'd really love your eyes on this MR before I assign to anyone else. See the notes in the description.

kevinmpowell:
@joesnellpdx I'd recommend adding a different image besides the placeholder 800x800 gray. A photo or something else will help show how the cropping will work with this technique.

Something along these lines...but maybe not this for copyright reasons:
screen shot 2019-02-22 at 3 22 55 pm

library/components/media/media.njk
```diff
@@ -0,0 +1,34 @@
+{% macro objectfitimage(
```
kevinmpowell: I would rename this file to `image` and make the macro match. I think this is a great "image as a component" with lots of functionality.

library/components/media/media.njk
```diff
@@ -0,0 +1,34 @@
...
+    container_class="spirit-object-fit",
```
kevinmpowell: I don't follow the need for the `container_class` here which makes this macro feel more like a mixin and less of a peer component. I see this as a peer to card, not something that needs to be masked:
```
<div class="spirit-card">
  <div class="spirit-image spirit-image--standard">
    <img class="spirit-image__inner" src="han.jpg"/>
  </div>
</div>
```

I think this will save on custom CSS for each component that consumes this as well.

kevinmpowell: NOTE ^ Don't take my code example above verbatim, I'm just illustrating the separation I'd expect for any composite component.

joesnellpdx:
@kevinmpowell Thanks for the review! Great advice as I was somewhere in between component and helper. 

Requesting another pass.

library/components/image/image.njk
```diff
@@ -0,0 +1,31 @@
...
+      {% endif %}
+      <figure class="spirit-image" itemprop="image" itemtype="http://schema.org/ImageObject">
+        <span class="spirit-image__image-wrap {{" " + type_class if type_class}}">
+          {% if img_src_alt %}
+            <img class="spirit-image__image {{img_class_alt}}" src="{{ img_src_alt }}" alt="{{ img_alt_text }}">
+          {% endif %}
+          <img class="spirit-image__image" src="{{ img_src }}" alt="{{ img_alt_text }}">
+        </span>
+        {% if img_caption %}
+          <figcaption class="spirit-caption">{{img_caption}}</figcaption>
```
kevinmpowell: Shouldn't this be `spirit-image__caption` ?

library/components/list/list.njk
```diff
@@ -41,7 +42,12 @@
             {% if href %}
               <a class="spirit-list-item__link-image" href="{{href}}">
             {% endif %}
-              <img class="spirit-list-item__image" src="{{image_src}}">
+              {{ image(
+                container_class="spirit-list-item",
```
kevinmpowell: This seems like an outdated invocation. `image()` doesn't have `container_class` as part of its API anymore right?

library/docs/sink-pages/components/images.njk
```diff
@@ -0,0 +1,105 @@
...
+            {% call library.container() %}
+                <div class="image-sink-test--narrow">
+                    <h2>
+                       Object Fit Images
+                    </h2>
+                    <h3>
+                       Portrait
+                    </h3>
+                    {{ library.image(
+                        container_class="sink-test",
```
kevinmpowell: I think all these `container_class`'s need to be dropped.

library/tokens/tokens.yaml
```diff
@@ -288,7 +288,12 @@ font :
         normal:        1.231
         breakpoint-l : 1.25
 
-
+ratio:
```
kevinmpowell: Ooh, I like these as tokens. Very nice.

joesnellpdx: Thanks!

joesnellpdx:
Adding @ismaeltoure7 and @lauralvictor to CR

See MR description for testing guidance.

joesnellpdx:
@lauralvictor 

- link one: should see peek-a-boo of cards in carousel left and right. Do the images look correct? Something else I'm missing?
- link two: this is an example of how the list items are not resilient if squished into 4up - this breaking is intentional. List items should not be in a 4up grid. Do the images look correct?

lauralvictor:
@joesnellpdx Yes, the images do look correct in both cases.

Issue #366, opened 2019-02-27
Tokens: update doc site tokens to match latest updates


joesnellpdx:
Verify tokens are accurate on doc site - in particular - new image ratio.

joesnellpdx:
PR here:
https://github.com/JDRF/spirit/pull/367

joesnellpdx:
Approved. Merged. Closing.

PR #367, opened 2019-02-27
feature/#366-tokens-update-doc:

joesnellpdx:
- add image ratio tokens to doc site
- add alignment helper classes to long form text doc page

Issue:
https://github.com/JDRF/spirit/issues/366

To test in doc site:
`/tokens.html#image-ratio` - see image ration tokens
`/components/long-form-text.html#text-alignment` - see text alignment options

### DIFF
--- a/doc-site/docs/components/long-form-text.njk
+++ b/doc-site/docs/components/long-form-text.njk
@@ -476,4 +476,183 @@
       examples=[ { example: lft_example_blockquote_center } ]
    ) }}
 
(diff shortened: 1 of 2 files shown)


ismaeltoure7:
@joesnellpdx It looks great! One question for the team about the aspect ratios, would it be more helpful to list the aspect ratios (3:4, 1:1, etc) instead of percentages under `value`? @jamesmelzer @lauralvictor

PR #369, opened 2019-03-06
101-list-pub

jamesmelzer:


doc-site/docs/components/lists.njk
```diff
@@ -0,0 +1,397 @@
...
+    }}
+
+  {% filter markdown %}
+      #### Item Caption 
+      Use item caption to summarize the item. This is useful for lists of articles or press releases. 
+  {% endfilter %}
+
+    {% set component_example %}
+      {% call library.list() %}
+          {{ library.list_item(title_text=" ", caption="JDRF, the leading global organization funding type 1 diabetes (T1D) research, celebrates Sherry and Larry Benaroya’s noteworthy contributions as volunteer supporters with the Distinction in Innovation Award.") }}
```
joesnellpdx: @jamesmelzer 

I don't think we want empty `title_text` here. Should use `Frederick Banting` to maintain structure as `Item Metadata`.

doc-site/docs/components/lists.njk
```diff
@@ -0,0 +1,397 @@
...
+      code_snippet_source=(component_example)
+      )
+    }}
+
+  {% filter markdown %}
+
+
+
+      ### Action links
+      Invite action with one or two links per list item, positioning underneath the item or to the right of the title.
```
joesnellpdx: @jamesmelzer 
Do we want to define that the first link should match the title link? Or, is that defined well enough in the editorial section?

doc-site/docs/components/lists.njk
```diff
@@ -0,0 +1,397 @@
...
+
+    {{ esds_doc.code_example_pair(
+      examples=[ { example: component_example } ],
+      code_snippet_source=(component_example)
+      )
+    }}
+
+  {% filter markdown %}
+      #### Item Caption 
+      Use item caption to summarize the item. This is useful for lists of articles or press releases.
```
joesnellpdx: @jamesmelzer do we want to emphasize limited text length here or maybe to the editorial section? CC @ismaeltoure7

doc-site/docs/components/lists.njk
jamesmelzer: Editorial covers it, but you're right: the code examples imply it but aren't clear enough. I'll add a sentence.

doc-site/docs/components/lists.njk
ismaeltoure7: @joesnellpdx @jamesmelzer It makes sense for it to be listed here.

jamesmelzer: @joesnellpdx Do we have a formal limit?

PR #376, opened 2019-03-22
Add page shell doc page

kevinmpowell:
Added Page Shell Doc Page


joesnellpdx:
@kevinmpowell 

I added "role" to the landmarks in the home page MR:
https://github.com/JDRF/spirit/pull/377/commits/dfc670c7ed97470108aa604af0406707656c6731

I don't think this affects what is transmitted in the doc site, unless we want to add something about the use of landmark roles.

https://www.w3.org/TR/wai-aria-practices/examples/landmarks/main.html

CC: @ismaeltoure7

PR #380, opened 2019-04-02
Feature/#379 flip hamburger nav

joesnellpdx:
@lauralvictor @ismaeltoure7 

Some quick updates before we release - 

- I moved the hamburger nav fly out to the right side - was discussed previously and not completed.
- There is a js error with esds on IE11 - this is known and not in our scope to fix at this time
- Fixed icon xbrowser issues - in xbrowser testing the nav, I found the feather icons and our custom icons did not play nice on IE11 - so I refactored all the feather icons (changed strokes to paths) so they all worked the same everywhere - fixed

have to gulp in library
for doc site ( after library compiled ):
-- npm run unlink-local-library
-- npm run link-local-library
-- gulp

Please review.


joesnellpdx:
@lauralvictor GREAT catch - please retest - only need to pull branch and gulp on doc site (if you have not done anything else with library since)

joesnellpdx:
@ismaeltoure7 

That is the 'focus' state. For accessibility reasons, there _should_ be visual focus when opening a component by WCAG 2.0 guidelines. You can see an example on Microsoft at the link below. Shrink the page down to view the hamburger - then open the nav and see the dotted line:
https://www.microsoft.com

I want to try and engineer everything for best practices, but understand if JDRF (you) want to override for design considerations. If you want me to remove any visual focus state in the hamburger nav when opened, I will. 

What is happening: 
- Hamburger is pressed
- User is 'moved' into the nav with the first item 'focused.' 
- If the user tabs through the nav - the will wind up back at the hamburger. 

It can all be done with no mouse and a user with disabilities can always see where they are. 

Just say the word to remove visual focus when opening the nav - and it will be done. 

CC: @patrickmckowen

ismaeltoure7:
@joesnellpdx I respect the application of best practices. However, I suggest removing the focus in this case and revisiting it as part of a bigger navigation refinement exercise, including .org.

joesnellpdx:
@ismaeltoure7 Removed. Please confirm we are :+1:

lauralvictor:
> @lauralvictor GREAT catch - please retest - only need to pull branch and gulp on doc site (if you have not done anything else with library since)

Thanks, I've re-tested and can now view the entire menu via scrollbar. https://www.evernote.com/shard/s403/sh/29083625-83b0-4296-9545-6f18037fb660/466409533e9425e3a00d13c652e6b1e3

PR #397, opened 2019-12-10
update font weights|sizes for labels add line height token

rafh:
This MR corresponds with the following ticket in JIRA: https://jdrf.atlassian.net/browse/CMEG-563

To test:
In branch and in the library:

grunt
go to `/sink-pages/components/forms.html`
view section with input|textareas - font for labels should be as follows:
- weight: 500
- size: 16px
- line-height: 1.3/20px

view section with input|textareas - font should be as follows:
- weight: 400

To test in doc-site:

be in this branch
in doc-site directory
npm run link-local-library
gulp
view `/components/forms.html#input`

### DIFF
--- a/library/components/form/form.scss
+++ b/library/components/form/form.scss
@@ -42,7 +42,7 @@ $spirit-form-element-max-width:             none;
 $spirit-form-element-max-width-100:         100%;
 
 // line height
-$spirit-form-label-line-height:             1.3;
+$spirit-form-label-line-height:             1.25;
 
 // space
 $spirit-form-element-padding:               10px 12px;
--- a/library/components/form/form.scss
+++ b/library/components/form/form.scss
@@ -41,6 +41,9 @@ $spirit-form-element-height:                48px;
 $spirit-form-element-max-width:             none;
 $spirit-form-element-max-width-100:         100%;
 
+// line height
+$spirit-form-label-line-height:             1.3;
+
 // space
 $spirit-form-element-padding:               10px 12px;
 $spirit-form-element-stack:                 $spirit-space-stack-half-x;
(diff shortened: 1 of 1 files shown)


library/components/form/form.scss
```diff
@@ -284,7 +284,8 @@ $spirit-textarea-placeholder-color:   $spirit-form-element-placeholder-color;
 // <label>
 $spirit-label-color:        $spirit-form-label-color;
 $spirit-label-font-family:  $spirit-form-font-family;
-$spirit-label-font-size:    $spirit-font-size-s;
+$spirit-label-font-size:    $spirit-font-size-m;
+$spirit-label-line-height:  $spirit-font-line-height-heading-level-3-normal;
```
patrickmckowen: @rafh  It's not obvious to me that I would use a token with the word "heading" in the name when styling a form label.  And the design calls for a value of `.8` and `heading-level-3` is `1.33`

library/components/form/form.scss
```diff
@@ -41,6 +41,9 @@ $spirit-form-element-height:                48px;
 $spirit-form-element-max-width:             none;
 $spirit-form-element-max-width-100:         100%;
 
+// line height
+$spirit-form-label-line-height:             1.3;
```
patrickmckowen: @rafh shouldn't this be `.8` instead of `1.3` to achieve the 20px line-height?

rafh: In order to achieve the 20px line-height, we will need to use 1.3.  This is due to us changing the label's font size to 16px. In order to get 20px exactly, it would technically need to be 1.25.  However, not all browsers round numbers the same.  To avoid this discrepancy I usually stop at the tenths place with integers.

rafh: @patrickmckowen I have updated the line-height value to be 1.25.  If you open the dev tools and replace this value with 20px, you will see that these two values are equal.

PR #406, opened 2020-01-08
added fixed left navigation

rafh:
[Original Ticket](https://jdrf.atlassian.net/browse/DGST-3)

The right navigation now persists as the user scrolls through the doc site.  I also moved the footer to prevent it from overlapping the navigation.


rafh:
I think this change gives the navigation some breathing room.

PR #408, opened 2020-01-09
feature/DGST-11-input-button-doc

joesnellpdx:
Add input button to forms page in doc site

Ticket: https://jdrf.atlassian.net/browse/DGST-11

`pull branch`
in doc site, run `gulp`
go to `/forms/` and view the input section (top) - now has input with button component

PR #409, opened 2020-01-10
Feature/dgst 12 add form password input doc

joesnellpdx:
This update has changes to both the library and the doc site

Library:
- update the link macro so it displays `"` correctly when used in the code example pairs on the doc site
- enable the use of `role` in the link macro for use on the doc site
- update the forms.js script for the password toggle to prevent default link action and `aria-clicked` attributes
- general code cleanup

Doc Site:
- Add content for password input with toggle 
- Add code example pair for password input with toggle

To view on doc site:
- pull branch
- in library directory run `gulp`
- switch to doc-site directory
- run `npm run link-local-library`
- run `gulp`
- go to `/components/forms.html#password-input-with-toggle` and review

### DIFF
--- a/doc-site/docs/components/forms.njk
+++ b/doc-site/docs/components/forms.njk
@@ -9,7 +9,6 @@
 
 {% block page_shell_content %}
 
-
   {% filter markdown %}
       ## Input
       Use input to gather a short string of text or a number. Also known as a text box.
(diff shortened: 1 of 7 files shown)


joesnellpdx:
> @joesnellpdx No feedback except for the subhead / sidebar nav labels and order.
> 
> I'm wondering if it should be:
> 
> -Input
> --Masks
> --With Button
> --With Password Toggle

@patrickmckowen I've pushed an update to update:
- Masks
- With Password Toggle

When this branch is approved and merged into develop, I will align with With Button before that branch is complete: https://github.com/JDRF/spirit/pull/408

I've also updated that ticket accordingly:
https://jdrf.atlassian.net/browse/DGST-11

PR #415, opened 2020-01-28
Feature/dgst 28 icon button enhancements

joesnellpdx:
This MR adds the following updates for icon button enhancements:
- add icon button sink
- update icon sizes, styles
- update njk button markup output + parameter options
- remove some 'spirit-icon' styles in button as can now inherit primary 'spirit-button' styles
- use common icon sizes in buttons
- add class / style for spirit-icon left and right
- Chose NOT to use a span around the label / button text by default to maintain current markup as much as possible. ISSUE - when text wraps, it is centered.
- Buttons with only icon (i.e. success) - changed icon size to match icon sizes used in designs for different button types (made larger)

Additional updates included:
- fix button groups sink repeating component ids
- add button group z-index when in use to allow border to show

Notes / Questions:
- Spirit Heading Classes in Long Form Text: spirit heading classes (i.e. `spirit-h4`) are ignored in `.spirit-long-form-text`. Should this be updated? New ticket?

To review:
1. Pull branch
2. in `/library` run `grunt`
3. Visit `/sink-pages/components/button-icons.html` and review
4. Visit `/sink-pages/components/buttons.html` and review
5. Visit `/sink-pages/components/button-groups.html` and review

### DIFF
--- a/library/components/button/button.scss
+++ b/library/components/button/button.scss
@@ -31,7 +31,7 @@ $spirit-button-icon-margin-bottom-small: -2px;
 $spirit-button-icon-margin-top-small: -2px;
 
 // Is Icon Button
-$spirit-button-is-icon-width: 130px;
+$spirit-button-is-icon-min-width: 130px;
 $spirit-button-is-icon-padding-default: 12px;
 $spirit-button-is-icon-padding-xs: 6px;
 $spirit-button-is-icon-padding-s: 10px;
@@ -115,8 +115,8 @@ $spirit-button-is-icon-padding-l: 20px;
   }
 
   .spirit-button__default-text:not([aria-hidden='true']) {
-    .spirit-icon--left,
-    .spirit-icon--right {
+    .spirit-button__icon--left,
+    .spirit-button__icon--right {
       visibility: visible;
     }
   }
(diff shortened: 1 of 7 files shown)


library/components/button/button.scss
```diff
@@ -24,22 +24,38 @@ $spirit-space-inset-squish-3-x-negative: -12px -24px !default;
...
 $spirit-space-inset-squish-6-x-negative: -24px -48px !default;
 
+// Icon Spacing
+$spirit-button-icon-margin-bottom-default: -3px;
+$spirit-button-icon-margin-top-default: -5px;
+$spirit-button-icon-margin-bottom-small: -2px;
+$spirit-button-icon-margin-top-small: -2px;
+
+// Is Icon Button
+$spirit-button-is-icon-width: 130px;
```
patrickmckowen: Minimum width?

library/components/button/button.scss
```diff
@@ -98,6 +114,29 @@ $spirit-space-inset-squish-6-x-negative: -24px -48px !default;
...
+
+  [class*='spirit-icon-'] {
+    flex: 0 0 $spirit-size-icon-m;
+    height: $spirit-size-icon-m;
+    margin-bottom: $spirit-button-icon-margin-bottom-default;
+    margin-top: $spirit-button-icon-margin-top-default;
+    width: $spirit-size-icon-m;
+  }
+
+  .spirit-icon--left {
```
patrickmckowen: Because this relies on the negative top/bottom margins, should we rename to be more specific to button?

joesnellpdx:
@patrickmckowen 

Review latest updates.

- added classes: `spirit-button__icon`, `spirit-button__icon--left` and `spirit-button__icon--right` 
- changed the variable title to `spirit-button-is-icon-min-width`
- moved the toggle buttons and made sure each has unique id
- scss updates - removed unnecessary specificity on the new icon classes 

Ideally, move the current styles for `spirit-icon` when inside a `spirit-button` to the new `spirit-button__icon` class, but that would be a breaking change and I don't think it is critical. Wanted to make sure to point that out. We _could_ have the style in two places and remove from `spirit-icon` in ver `2.0.0`, but I lean towards leaving as is.

